PR #13, opened 2024-08-11
convert foam dictionary to neofoam dictionary

HenningScheufler:
added feature:

- [x] conversion foam dictionary to neofoam dictionary
- [x] add launch json to simlify debugging in vscode


HenningScheufler:
@bevanwsjones

i already merged it into stack dsl

PR #17, opened 2024-09-24
use branch with new field ctr

greole:
This PR uses the constructor to construct a field from existing memory https://github.com/exasim-project/NeoFOAM/pull/145

Additional:
- fix CI pipeline


.github/workflows/build.yaml
```diff
@@ -40,9 +40,10 @@ jobs:
      - name: Add repos
        run: |
          sudo add-apt-repository 'deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-16 main'
-         wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | sudo bash
```
HenningScheufler, line 43: Why doesn't this work

include/FoamAdapter/readers/foamFields.hpp
```diff
@@ -17,31 +17,10 @@ template<typename FoamType>
...
-    // Create device-side views
-    type_container_t nfField(exec, field.size());
-
-    type_container_t CPUField(NeoFOAM::CPUExecutor {}, field.size());
-    Kokkos::View<type_primitive_t*, Kokkos::HostSpace, Kokkos::MemoryUnmanaged> CPU_view(
-        CPUField.data(), field.size()
+    using mapped_t = typename type_map<FoamType>::mapped_type;
+    type_container_t nfField(
+        exec, reinterpret_cast<const mapped_t*>(field.cdata()), static_cast<size_t>(field.size())
     );
```
HenningScheufler, line 23: Why reinterpret cast?

HenningScheufler (approved):

.github/workflows/build.yaml
greole: I dont know  :/ . I would fix it like this for now. If we find a better way we can use wget again. I assume esi changed something on their side, since it suddenly stopped working.

include/FoamAdapter/readers/foamFields.hpp
greole: cdata can be a pointer to foam::vector, in which case we have to cast foam::vector to neofoam::vector.

PR #29, opened 2024-11-23
Discussion about clang format

HenningScheufler:
how to apply change to the whole project
```
find . -name '*.cpp' -o -name '*.hpp' | xargs clang-format -i
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -21,9 +21,9 @@ AlignEscapedNewlines: Right
 # Align operands after operators (+,*,<<) (see BreakBeforeBinaryOperators)
 AlignOperands: AlignAfterOperator
 AlignTrailingComments: true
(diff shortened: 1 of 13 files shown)


HenningScheufler:
@greole

greole:
I would rather say we use the same pre-commit based approach as in NeoFOAM. To run clang-format and all other checks regardless of a git commit would be just a `pre-commit run --all`

greole:
> how to apply change to the whole project
> 
> ```
> find . -name '*.cpp' -o -name '*.hpp' | xargs clang-format -i
> ```

I would rather say we use the same pre-commit based approach as in NeoFOAM. To run clang-format and all other checks regardless of a git commit would be just a pre-commit run --all

greole (approved):
LGTM! But the corresponding check is failing.

PR #30, opened 2024-11-24
FIX: segfault writing nf field

HenningScheufler:
Writing a GPU field result in a seg fault as the memory on the GPU was accessed instead of the host field


HenningScheufler:
@greole

greole (approved):

PR #34, opened 2024-11-27
Draft: Feat/dsl implementation

HenningScheufler:
This is the most uptodate branch (only compiles with clang)

@gregorweiss 
@greole


HenningScheufler:
wrong branch

PR #44, opened 2024-12-30
Benchmark/scalar advection

HenningScheufler:
Adds to new example scalarFoamAdvection and scalarNeoFoamAdvection

and a very unsophisticated benchmark for discussion and to try it out

The goal of NeoFOAM is to reduce the clockTime of solvers while ensuring the same results. The implementation of a Foam and NeoFOAM version side by side would enables us to understand the performance impact and also guides a way to a clean syntax

Features:

* the CFL number is always on the CPU (preliminary solution turn it off)
* 

This benchmarking is quite tedious and should be automated  with snakemake to do quick parameter scans.

What are your thoughts about a benchmarking with this approach and automating the process as soon as possible?

@gregorweiss @greole @bevanwsjones 

Note: The Serial and the OpenFOAM implementation of NeoFOAM show a similar performance

### DIFF
--- a/examples/scalarAdvection/scalarFoamAdvection.cpp
+++ b/examples/scalarAdvection/scalarFoamAdvection.cpp
@@ -78,9 +78,12 @@ int main(int argc, char* argv[])
         // Info << "max(U) : " << max(U).value() << endl;
 
         // advance Foam fields in time
-        Foam::fvScalarMatrix TEqn(fvm::ddt(T) + fvc::div(phi, T));
+        {
+            addProfiling(solveT, "solveT");
+            Foam::fvScalarMatrix TEqn(fvm::ddt(T) + fvc::div(phi, T));
 
-        TEqn.solve();
+            TEqn.solve();
+        }
 
         if (runTime.outputTime())
         {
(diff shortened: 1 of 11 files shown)


examples/scalarAdvection/scalarFoamAdvection.cpp
```diff
@@ -0,0 +1,101 @@
...
+// SPDX-FileCopyrightText: 2023 NeoFOAM authors
+
+#include "FoamAdapter/FoamAdapter.hpp"
+#include "NeoFOAM/dsl/expression.hpp"
+#include "NeoFOAM/dsl/solver.hpp"
+#include "NeoFOAM/dsl/ddt.hpp"
+#include "FoamAdapter/readers/foamDictionary.hpp"
+
+#include "NeoFOAM/dsl/implicit.hpp"
+#include "NeoFOAM/dsl/explicit.hpp"
```
greole, line 11:
```suggestion
#include "FoamAdapter/readers/foamDictionary.hpp"

#include "NeoFOAM/NeoFOAM.hpp"
```

greole:
> This benchmarking is quite tedious and should be automated with snakemake to do quick parameter scans.

If there are no hard objections i would propose to use [obr](https://github.com/exasim-project/OBR).

PR #71, opened 2025-04-20
updated cmake files to build also with petsc solver

happyKitesurfer:
I updated the cmake files in order that the FoamAdapter can be build also with petsc

### DIFF
--- a/examples/heatTransfer/CMakeLists.txt
+++ b/examples/heatTransfer/CMakeLists.txt
@@ -5,3 +5,10 @@
 add_executable(heatTransfer heatTransfer.cpp)
 
 target_link_libraries(heatTransfer FoamAdapter)
+
+if(NeoN_WITH_PETSC)
+  target_link_libraries(heatTransfer ${PETSc_LIBRARIES} fftw3 mpi)
+  target_include_directories(heatTransfer PUBLIC ${PETSc_INCLUDE_DIRS})
+  target_link_directories(heatTransfer PUBLIC ${PETSc_LIBRARY_DIRS})
+  link_directories(${PETSc_LIBRARY_DIRS})
+endif()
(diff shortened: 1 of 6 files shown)


examples/heatTransfer/CMakeLists.txt
```diff
@@ -5,3 +5,10 @@
 add_executable(heatTransfer heatTransfer.cpp)
 
 target_link_libraries(heatTransfer FoamAdapter)
+
+if(NeoN_WITH_PETSC)
```
greole, line 9: I suspect that this should actually be part of neons public API, i.e. if neon is build with PETSc support the neon_public_api target should introduce this dependency.

happyKitesurfer: That's actually strange. When I do not add this, the linker complains about undefined references even though neon is build with petsc. Actually the symbols should be defined in some NeoN libraries. I'll check again

happyKitesurfer: ok I solved the issue https://github.com/exasim-project/NeoN/pull/327

happyKitesurfer: needed to add

`if(NeoN_WITH_PETSC)
  add_definitions(${PETSc_DEFINITIONS})
  target_link_libraries(NeoN_public_api INTERFACE ${PETSc_LIBRARIES})
  target_link_libraries(NeoN_public_api INTERFACE ${PETSc_LIBRARIES}
                                                  ${PETSc_LIBRARY_DIRS}/libkokkoscore.so mpi)
  target_include_directories(NeoN_public_api INTERFACE ${PETSc_INCLUDE_DIRS})
  target_link_directories(NeoN_public_api INTERFACE ${PETSc_LIBRARY_DIRS})
  link_directories(${PETSc_LIBRARY_DIRS})
  target_compile_definitions(NeoN_public_api INTERFACE NF_WITH_PETSC=1)
else()`

in include/CMakeLists.txt

PR #89, opened 2025-08-07
Use a project access token for mirroring

chihtaw:
Use a project access token instead of a personal access token for mirroring the GitHub repo to LRZ GitLab 
Then the mirroring job will not tied to a personal account on LRZ GitLab.


greole (approved):
LGTM!

PR #90, opened 2025-08-18
Add a config file for running GitLab CI pipelines on TUM COMA cluster

chihtaw:
The CI pipeline should be triggered by any changes to FoamAdapter and do the following:

1. Clone NeoN main branch
2. Build FoamAdapter along with the main branch of NeoN.
3. Run unit tests of FoamAdapter


chihtaw:
I think we can close this PR because #91 does the same thing in a single job and we have problem with uploading artifacts of large size.

PR #93, opened 2025-08-21
Fetch Catch2 and make it available to benchmarks

chihtaw:
The config step fails for the CMake preset `profiling` because Catch2 cannot be found for linking targets in benchmarks directory.


greole (approved):
LGTM!

PR #100, opened 2025-09-04
Disable the building of benchmarks in the triggered CI job

chihtaw:
The current triggered CI job  is used to check if changes in NeoN work for FoamAdapter. For this purpose, it builds FoamAdapter's tests. In addition, it also builds FoamAdapter benchmarks. However, we do not need benchmarks in the CI job.


greole (approved):
LGTM!

PR #101, opened 2025-09-04
Skip changelog check job for a PR whenever skip-changelog label is present

chihtaw:
The original static checks only run for an opened PR or new commits. A reopened PR is not considered, but should also be checked.

Originally, the changelog check is included in the static checks. However, the types of PRs for the changelog check can be different from the types of PRs for static checks: the changelog check workflow can also be triggered by **a label change** for an opened PR. Therefore, the changelog check should be defined in a separate workflow.

Whenever the `skip-changelog` label is present, the changelog check will be skipped.

### DIFF
--- a/.github/workflows/changelog_check.yaml
+++ b/.github/workflows/changelog_check.yaml
@@ -1,0 +1,17 @@
+name: Changelog check
+run-name: Changelog check
+
+on:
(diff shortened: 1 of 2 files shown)


greole (approved):
LGTM!

PR #107, opened 2025-09-19
Set up CI on AMD GPU

chihtaw:
## The current status
**The GitHub workflow of triggering LRZ GitLab CI**:
- The configuration file is confusing because the filename `gitlab-ci.yaml` is similar to the configuration file for the LRZ GitLab CI `.gitlab-ci.yml`.
- The URLs are too long to read.
- The GitHub workflow continues to query the status of a triggered LRZ GitLab CI pipeline even if the pipeline failed to finish.

**The LRZ GitLab CI**:
- The CI setup only supports NVIDIA GPUs.
- Some code duplication exists for the build-and-test jobs which are triggered from either FoamAdapter GitHub or NeoN GitHub.
- All the OpenFOAM environment variables are defined globally.
- The version of OpenFOAM is defined together with all other OpenFOAM environment variables.

## The new status
**The GitHub workflow of triggering LRZ GitLab CI**:
- The configuration file for the GitHub workflow is renamed `trigger-lrz-gitlab-ci.yml` to avoid the confusion and to be clearer what the workflow does.
- The URLs are shortened by introducing environment variables as shell variables in place of GitHub variables.
- The GitHub workflow is stopped as soon as a failed status of the triggered CI pipeline is detected.

**The LRZ GitLab CI**:
- A new docker image with OpenFOAM installed is prepared for AMD GPUs.
- The CI setup now supports both NVIDIA and AMD GPUs.
- The configuration file is highly modularized so that different build-and-test jobs are defined by including required components.
- The definition of the OpenFOAM environment variables is not global anymore and is made reusable. It makes the intention explicit by including the required OpenFOAM component in jobs.
- The definition of the OpenFOAM version is defined together with the chosen docker image, which makes it easier to switch to a different version of OpenFOAM depending on the chosen docker image.
- A shell script is prepared for all the build-and-test jobs regardless of the trigger sources and the GPU vendors.


.github/workflows/trigger-lrz-gitlab-ci.yaml
```diff
@@ -14,19 +14,20 @@ concurrency:
...
 
 jobs:
-  trigger-lrz-ci:
+  build-and-test:
     name: Trigger and run CI on GPU (${{ github.event.repository.name }})
     runs-on: ubuntu-latest
-    timeout-minutes: 1440   # 24 hours max runtime
+    timeout-minutes: 1440 # 24 hours
 
-    # Skip if PR has Skip-build label
```
greole, line 22: I think you can leave that comment in.

ci/build-and-test.sh
```diff
@@ -0,0 +1,75 @@
+# SPDX-FileCopyrightText: 2023 - 2025 NeoN authors
+#
+# SPDX-License-Identifier: Unlicense
+
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Argument parsing
+GPU_VENDOR=${1:?Error: GPU vendor (nvidia|amd) must be specified}
+NEON_BRANCH=${2:-main} # Default to 'main' if not provided
```
greole, line 10: Nice! Never seen that before, is that standard bash?

greole (approved):
LGTM!

ci/build-and-test.sh
chihtaw: Yes, they are standard bash parameter expansions.

PR #109, opened 2025-10-01
Add benchmark data

dheerajraghunathan:
Added script to perform benchmarking from the NeoFOAM-BenchmarkData repository.


chihtaw:
Additionally, the benchmark job on GitHub CI should depends on the build-and-test job. That is, we do benchmarking only after the current branch passes all the tests.

PR #114, opened 2025-10-07
Enh/unify solver

greole:


benchmarks/bench_dsl.cpp
```diff
@@ -22,29 +22,31 @@ namespace dsl = NeoN::dsl;
...
+
+    Foam::surfaceScalarField ofGamma(
+        Foam::IOobject("Gamma", "0", mesh, Foam::IOobject::NO_READ, Foam::IOobject::AUTO_WRITE),
+        mesh,
+        Foam::dimensionedScalar("phi", Foam::dimensionSet(0, 2, -1, 0, 0), 0.0)
+    );
 
     SECTION("OpenFOAM")
     {
         std::unique_ptr<Foam::fvMesh> meshPtr = FoamAdapter::createMesh(runTime);
```
chihtaw:
```suggestion
        std::unique_ptr<Foam::fvMesh> meshPtr = nf::createMesh(runTime);
```

benchmarks/bench_dsl.cpp
```diff
@@ -12,7 +12,7 @@
 #include "common.hpp"
 
 namespace nnfvcc = NeoN::finiteVolume::cellCentred;
```
chihtaw:
```suggestion
namespace fvcc = NeoN::finiteVolume::cellCentred;
```
We need to be more consistent with the naming of the namespace. 
We go with either `nnfvcc` or `fvcc`. I would prefer the shorter one.

If we choose `fvcc`, then all the occurrences of `nnfvcc` in the same file should be replaced with `fvcc`.

benchmarks/bench_dsl.cpp
```diff
@@ -100,45 +97,25 @@ TEST_CASE("advection–diffusion-equation_scalar")
...
-        Foam::surfaceScalarField ofGamma(
-            Foam::IOobject("Gamma", "0", mesh, Foam::IOobject::NO_READ, Foam::IOobject::AUTO_WRITE),
-            mesh,
-            Foam::dimensionedScalar("Gamma", Foam::dimensionSet(0, 2, -1, 0, 0), 1.0)
-        );
-
-        auto nfGamma = FoamAdapter::constructSurfaceField(exec, nfMesh, ofGamma);
-
-        NeoN::Dictionary fvSolutionExpDict;
+        auto nfPhi = FoamAdapter::constructSurfaceField(rt.exec, rt.nfMesh, ofPhi);
```
chihtaw:
```suggestion
        auto nfPhi = nf::constructSurfaceField(rt.exec, rt.nfMesh, ofPhi);
```

benchmarks/bench_dsl.cpp
```diff
@@ -100,45 +97,25 @@ TEST_CASE("advection–diffusion-equation_scalar")
...
+        auto nfGamma = FoamAdapter::constructSurfaceField(rt.exec, rt.nfMesh, ofGamma);
```
chihtaw:
```suggestion
        auto nfGamma = nf::constructSurfaceField(rt.exec, rt.nfMesh, ofGamma);
```

benchmarks/bench_dsl.cpp
```diff
@@ -22,29 +22,31 @@ namespace dsl = NeoN::dsl;
...
+    Foam::surfaceScalarField ofGamma(
+        Foam::IOobject("Gamma", "0", mesh, Foam::IOobject::NO_READ, Foam::IOobject::AUTO_WRITE),
+        mesh,
+        Foam::dimensionedScalar("phi", Foam::dimensionSet(0, 2, -1, 0, 0), 0.0)
+    );
 
     SECTION("OpenFOAM")
     {
         std::unique_ptr<Foam::fvMesh> meshPtr = FoamAdapter::createMesh(runTime);
         Foam::fvMesh& mesh = *meshPtr;
```
chihtaw: I found some redeclaration of the variable `mesh` like this in here and later in `SECTION("NeoN")`.
At the top level, there is already a declaration of `mesh`.

It causes variable shadowing. This shadowing is legal but is confusing and error-prone. 
It would be better to give locally created variables distinct names.

benchmarks/bench_dsl.cpp
```diff
@@ -148,39 +125,36 @@ TEST_CASE("advection–diffusion-equation_scalar")
...
-                    fvSolutionExpDict
+                    rt
                 );
-                advectDiffEqn.assemble(1.0, 1e-3);
+                advectDiffEqn.assemble();
                 return;
             };
         }
 
         SECTION(std::string("implicit"))
```
chihtaw:
```suggestion
        SECTION(std::string("implicit-time-integration"))
```

chihtaw (commented):
In benchmarks, we correct boundary conditions for NeoN, but do not do that for OpenFOAM.
Why?

greole:
> In benchmarks, we correct boundary conditions for NeoN, but do not do that for OpenFOAM.
Why?

For now we haven't decided to do it automatically.

PR #119, opened 2025-10-14
Add exp::grad(U) to neoIcoFoam

greole:
This PR is the adds `exp::grad(U)` to neoIcoFoam.

Additional:
- fix solver and stopping criterion mapping for new ginkgo version, since ginkgo now complains about extra config entries.

**Note**: I haven't fully validated the solver yet. But I think we should do it separately in order to get the pipeline up and running again.

### DIFF
--- a/src/compatibility/fvSolution.cpp
+++ b/src/compatibility/fvSolution.cpp
@@ -62,7 +62,7 @@ void updatePreconditioner(NeoN::Dictionary& solverDict)
          )},
     };
 
-    // if no
+    // if no preconditioner is set but smoother switch to BiCGStab with BJ
     if (!solverDict.contains("preconditioner"))
     {
         solverDict.insert("preconditioner", preconditionerMap["DIC"]);
--- a/examples/neoIcoFoam/neoIcoFoam.cpp
+++ b/examples/neoIcoFoam/neoIcoFoam.cpp
@@ -84,11 +84,7 @@ int main(int argc, char* argv[])
             auto& oldU = fvcc::oldTime(U);
             oldU.internalVector() = U.internalVector();
 
-            // FIXME TODO sync runTime
-            auto t = runTime.time().value();
-            auto dt = runTime.deltaT().value();
-
-            auto coNum = fvcc::computeCoNum(phi, dt);
+            auto coNum = fvcc::computeCoNum(phi, rt.dt);
             if (rt.adjustTimeStep)
             {
                 nf::setDeltaT(runTime, rt, coNum);
@@ -103,7 +99,7 @@ int main(int argc, char* argv[])
 
             if (piso.momentumPredictor())
             {
-                UEqn.solve(dsl::exp::grad(p));
+                UEqn.solve();
             }
             else
             {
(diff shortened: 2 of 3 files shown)


include/FoamAdapter/datastructures/expression.hpp
```diff
@@ -123,33 +125,42 @@ class PDESolver
...
-            //mapFvSolution(solverDict.get<NeoN::Dictionary>(psi_.name)),
+            fieldSolverDict,
             functs
         );
 
-        std::cout << __FILE__ << ":" << __LINE__ << " solver stats:\n"
-                  << "\t num iter: " << stats.numIter
-                  << "\n\t initial residual norm: " << stats.initResNorm
-                  << "\n\t final residual norm: " << stats.finalResNorm << std::endl;
+        std::cout << "[NeoN] Solving for " << psi_.name << ":"
```
HenningScheufler: Mark as todo for potential improvement with logging;

Will become an issue in parallel runs

src/compatibility/fvSolution.cpp
```diff
@@ -96,32 +111,46 @@ void updateCriteria(NeoN::Dictionary& solverDict)
         solverDict.insert("criteria", NeoN::Dictionary());
     }
 
-
-    NeoN::Dictionary& criteriaDict = solverDict.subDict("criteria");
-
     // set default max iteration count
-    criteriaDict.insert("iteration", 1000);
+    {
```
HenningScheufler: The solver interace is currently tailored to Ginkgo this needs to be wrapped in the future to also be able to map to PETSC

include/FoamAdapter/datastructures/expression.hpp
```diff
@@ -123,33 +125,42 @@ class PDESolver
...
+            functs =
+                needReference_
+                    ? std::vector<NeoN::dsl::PostAssemblyBase<ValueType>> {SetReference<ValueType>(
+                          pRefCell_,
+                          pRefValue_
+                      )}
+                    : std::vector<NeoN::dsl::PostAssemblyBase<ValueType>> {};
         }
 
         // FIXME TODO this will create the sparsity pattern and potentially the ls
```
chihtaw: If the `TODO` is finished, please update the comment.

src/compatibility/fvSolution.cpp
```diff
@@ -23,32 +23,35 @@ void updateSolver(NeoN::Dictionary& solverDict)
         {"PCG", {"Ginkgo", "solver::Cg"}},
         {"PBiCG", {"Ginkgo", "solver::Bicg"}},
         {"PBiCGStab", {"Ginkgo", "solver::Bicgstab"}},
+        {"smoothSolver", {"Ginkgo", "solver::Bicgstab"}},
```
chihtaw: Why is `smoothSolver` mapped to `BiCGStab`?

In OpenFOAM, smootthSolver can be other smoothers.

src/compatibility/fvSolution.cpp
```diff
@@ -96,32 +111,46 @@ void updateCriteria(NeoN::Dictionary& solverDict)
...
+    NeoN::Dictionary& criteriaDict = solverDict.subDict("criteria");
+    std::cout << __FILE__ << ":\n\tStopping criteria: " << criteriaDict << "\n";
 }
 
 
 NeoN::Dictionary mapFvSolution(const NeoN::Dictionary& solverDict)
 {
     NeoN::Dictionary modSolverDict = solverDict;
 
+    if (solverDict.contains("configFile")) return solverDict;
```
chihtaw:
```suggestion
    if (modSolverDict.contains("configFile")) return modSolverDict;
```
To avoid confusion from aliasing the input dictionray

PR #134, opened 2025-10-25
add grad(p)

greole:
In OpenFOAMs pressure velocity coupling procedures e.g. in `icoFoam`, `simpleFoam`, `pimpleFoam` and others, the velocity predictor is implemented as:

```
if (piso.momentumPredictor())
{
    solve(UEqn == -fvc::grad(p));
}
```
note that the operator== returns a temporary clone of the original `UEqn` including the new rhs `-fvc::grad(p)`. When solving for U the orginal `UEqn` wont include  -fvc::grad(p) and thus  rAU and HbyA neither.

This PR adds a new `solve(rhs)` operating only  on a temporary clone of the original `linearSystem`.

Additional changes and fixes:
- [x] rename expression.hpp to pdeSolver.hpp to match the class name
- [x] failing benchmark 
- [x] Add unit test comparing OF and NF solves


examples/neoIcoFoam/neoIcoFoam.cpp
greole, line 1: add unit test for it

benchmarks/bench_dsl.cpp
```diff
@@ -124,15 +126,13 @@ TEST_CASE("advection–diffusion-equation_scalar")
 
             BENCHMARK(std::string(execName))
             {
-                // Momentum predictor
-                nffvcc::Expression<NeoN::scalar> advectDiffEqn(
+                auto eqn = nf::PDESolver(
```
HenningScheufler, line 129: I think i missed that change but nf::PDESolver?

Why not PDE?

examples/neoIcoFoam/neoIcoFoam.cpp
```diff
@@ -99,7 +99,8 @@ int main(int argc, char* argv[])
 
             if (piso.momentumPredictor())
             {
-                UEqn.solve();
+                // NOTE solve on a temporary clone of UEqn
+                UEqn.solve(-1.0 * dsl::exp::grad(p));
```
HenningScheufler, line 104: The PISO algorithm requires that UEqn has no contribution from the pressure.

OpenFOAM solved this with a free function (solve) and by creating a new temporary matrix

include/FoamAdapter/datastructures/pdeSolver.hpp
```diff
@@ -136,14 +152,13 @@ class PDESolver
...
         }
 
-        // FIXME TODO this will create the sparsity pattern and potentially the ls
-        // again even if it has been created already
         auto solverDict = runTime_.fvSolutionDict.get<NeoN::Dictionary>("solvers");
         auto fieldSolverDict = solverDict.get<NeoN::Dictionary>(psi_.name);
-        auto stats = NeoN::dsl::detail::iterativeSolveImpl(
-            expr_,
+
+        auto stats = ::NeoN::dsl::detail::iterativeSolveImpl(
```
HenningScheufler: why ::Neon

include/FoamAdapter/datastructures/pdeSolver.hpp
```diff
@@ -119,8 +119,24 @@ class PDESolver
...
+        auto ls = NeoN::la::LinearSystem<ValueType, IndexType>(ls_);
+        expr.addOperator(-1.0 * rhs);
+        assemble();
+        return solveImpl(expr, ls);
+    }
+
+private:
+
+    NeoN::la::SolverStats solveImpl(
+             dsl::Expression<ValueType>& expr,
```
HenningScheufler, line 138: I assumed PDESolver is a new name for dsl::Expression?

benchmarks/bench_dsl.cpp
greole: So the main reason for renaming it from from `nf::expression` was that we had two classes named expression which are very different. Since the the `PDESolver` class gets a PDE in the form of an expression as input and provides a solve function I found the name adequate. I think PDE it self would be misleading because it is only part of the responsibility here, since it also stores everyting needed to solve the PDE.

examples/neoIcoFoam/neoIcoFoam.cpp
greole: I can add a free function like in OpenFOAM i was a bit hesitant to add it just here because i always found the `UEqn ==`  a source for confusion. We can discuss and add the operator== in later PR though.

include/FoamAdapter/datastructures/pdeSolver.hpp
greole: good catch, some artifact from debugging. Is now removed.

include/FoamAdapter/datastructures/pdeSolver.hpp
greole: See comment above, no they are different things. The dsl::Expression encodes a PDE. The PDESolver holds the expression and the runtime in order solve the expression.

test/test_pressureVelocityCoupling.cpp
```diff
@@ -219,6 +210,169 @@ TEST_CASE("PressureVelocityCoupling")
...
+            {
+                REQUIRE(hostnfU2.view()[celli][0] == Catch::Approx(ofU[celli][0]).margin(1e-12));
+                REQUIRE(hostnfU2.view()[celli][1] == Catch::Approx(ofU[celli][1]).margin(1e-12));
+                // NOTE we lower the criterion here because OF explicitly zeros in the 2D case
+                REQUIRE(hostnfU2.view()[celli][2] == Catch::Approx(ofU[celli][2]).margin(1e-06));
+            }
+
+            SECTION("HbyA modified U")
+            {
+                ofU.correctBoundaryConditions();
```
HenningScheufler, line 255: Add comment that solve update the U and nfU.

The test would be more isolated and explicit by modiffy U and nfU
e.g.

forAll(U,celli)
{
U[celli].x() = celli + 1
U[celli].y() = celli + 2
U[celli].z() = celli + 3
}

test/test_pressureVelocityCoupling.cpp
```diff
@@ -219,6 +210,169 @@ TEST_CASE("PressureVelocityCoupling")
...
+
+            Foam::solve(ofUEqn == -Foam::fvc::grad(ofp));
+
+            nf::PDESolver<NeoN::Vec3> nfUEqn(
+                dsl::imp::ddt(nfU) + dsl::imp::div(nfPhi, nfU) - dsl::imp::laplacian(nfNu, nfU),
+                nfU,
+                rt
+            );
+
+            nfUEqn.solve(-1.0 * dsl::exp::grad(nfP));
```
HenningScheufler, line 324: `Foam::solve(ofUEqn == -Foam::fvc::grad(ofp));`

This is a lot more explicit about copying compared to this

`nfUEqn.solve(-1.0 * dsl::exp::grad(nfP));`

Why not add a free function:

`solve(nfUEqn - dsl::exp::grad(nfP))`

However, this could also be done in a later stage

test/test_pressureVelocityCoupling.cpp
```diff
@@ -219,6 +210,169 @@ TEST_CASE("PressureVelocityCoupling")
...
+            {
+                REQUIRE(hostnfU2.view()[celli][0] == Catch::Approx(ofU[celli][0]).margin(1e-12));
+                REQUIRE(hostnfU2.view()[celli][1] == Catch::Approx(ofU[celli][1]).margin(1e-12));
+                // NOTE we lower the criterion here because OF explicitly zeros in the 2D case
+                REQUIRE(hostnfU2.view()[celli][2] == Catch::Approx(ofU[celli][2]).margin(1e-06));
+            }
+
+            SECTION("HbyA modified U")
+            {
+                ofU.correctBoundaryConditions();
```
HenningScheufler, line 344: comments that U is modified by solve or explicitly add a "loop that modifies U"

Note i see this as integration that 

`nfUEqn.solve(-1.0 * dsl::exp::grad(nfP));`

does not modify nfUEqn

test/test_pressureVelocityCoupling.cpp
greole: See my previous comment

> I can add a free function like in OpenFOAM i was a bit hesitant to add it just here because i always found the UEqn == a source for confusion. We can discuss and add the operator== in later PR though.

With confusing, I mean that anywhere else == is a boolean comparison.

In any case I'll leave a todo note about a free function.

HenningScheufler (approved):

test/test_pressureVelocityCoupling.cpp
HenningScheufler: Note: that the operator== can not be mapped in the python as the return value has to be a bool.

But we should capture it

PR #137, opened 2025-10-28
Add an action to update NeoN SHA in CMake and submodule

yhmtsai:
I understand FoamAdapter heavily relies on NeoN, so you try to make them work at the same time.
This PR is to make the job more clear, which raises error at right place.

Currently, NeoN has a github action to test whether FoamAdapter can run on top of the changes in NeoN.
It introduces the following issues when FoamAdapter must change something to fit the new NeoN
1. In theory, it requires both pr on both side are accepted and merged together
-> upstream library needs to wait for the downstream library sounds weird to me
2. In practice, FoamAdapter needs longer time and change after NeoN actually settle the change.
-> late merge in FoamAdapter seems to be reasonable and hurry for merging seems to be dangerous.
3. Before FoamAdapter updated, the action in NeoN keeps fails. We might need to ignore it several times
-> If we ignore the test several times, the test is not a test. And FoamAdapter are failed when user download it

I think frequently and automatically update NeoN version (SHA) in FoamAdapter might help in some senses.
1. NeoN does not need to know how to build FoamAdapter and wait for FoamAdapter check (simplify the CI setup in NeoN)
2. Specific SHA in FoamAdapter ensures that user can always get the working version from automatic download. (Note. it does not solve the issue when users install NeoN by themselves)
3. FoamAdapter failed test are located in FoamAdapter not NeoN.

It is similar to github dependabot.
Dependabot does not support CMake system, so I did it in this PR together with submodule
If this PR is merged, github action will automatically create a branch and corresponding PR to update NeoN SHA. It checks the NeoN version everyday. 
However, it does not close the old PR currently if there is a new update from NeoN.
If we change all test to rely on submodule not CPM, we can rely on dependabot. (I will also suggest that, but I do not check how it works on dependabot yet)
The automatic PR does not perform the CI automatically because of Github design.
Please look https://github.com/orgs/community/discussions/62346
It can be solved by using PAT or do some trigger event on that pr by reviewer.
The automatic PR can be checked in https://github.com/exasim-project/FoamAdapter/pull/145

In theory, I can make NeoN to create PR and branch in FoamAdapter, but I feel it unnecessary 

If this approach sounds reasonable, we can delete the FoamAdapter CI setup in NeoN.


yhmtsai:
test #209 and #210 in favor of less changes

PR #143, opened 2025-11-04
Automatic merging interdependent branches on NeoFOAM and NeoN 

chihtaw:
### Motivation

When a branch with breaking changes is introduced on NeoN, another branch from NeoFOAM is required to adapt to the changes in NeoN. If the NeoN's branch is merged manually before the NeoFOAM's branch is merged, the NeoFOAM will be broken.

There is a need of a mechanism to coordinate the merging of two branches so that the two branches with merged changes on both NeoFOAM and NeoN can still work together.

### Idea

- Both branches on NeoFOAM and NeoN which should work together need to be finished before merging.
- Both branches need to be merged at the same time (or in a very short time).

### Approach: Coordinated merge

**The procedure**:

1. Two feature branches on NeoFOAM and NeoN which should work together have to use the same branch name.
2. The two PRs for the two branches need to have target branches with the same branch name.
3. Both the two PRs need to be approved before merging.
4. Merging of the two PRs on both NeoFOAM and NeoN is performed automatically via a pre-defined GitHub workflow by adding the PR label `coordinated-merge`.

**Note**:
The automatic merging of the two PRs should be done in NeoFOAM because NeoFOAM is the downstream repo of NeoN. Merging should take place only when NeoFOAM's PR is done and ready to merge.

### DIFF
--- a/.github/workflows/coordinated-merge.yaml
+++ b/.github/workflows/coordinated-merge.yaml
@@ -85,8 +85,7 @@ jobs:
           gh pr merge "$neon_pr_number" \
             --repo $NEON_REPO \
             --merge \
-            --admin \
-            --delete-branch
+            --admin
           echo "NeoN PR merged."
 
       - name: Merge FoamAdapter PR
@@ -96,8 +95,7 @@ jobs:
           gh pr merge ${{ github.event.pull_request.number }} \
             --repo $FOAM_REPO \
             --merge \
-            --admin \
-            --delete-branch
+            --admin
           echo "FoamAdapter PR merged."
 
       # - name: Trigger LRZ GitLab CI
--- a/.github/workflows/coordinated-merge.yaml
+++ b/.github/workflows/coordinated-merge.yaml
@@ -1,0 +1,72 @@
+name: Coordinated Merge for Existing PRs (NeoN + FoamAdapter)
+
+on:
+  workflow_dispatch:
+    inputs:
+      neon_pr:
+        description: "NeoN PR number (e.g. 123)"
+        required: true
+      foam_pr:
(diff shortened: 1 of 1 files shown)


.github/workflows/coordinated-merge.yaml
```diff
@@ -0,0 +1,114 @@
...
+    types: [labeled, unlabeled, synchronize, reopened, opened]
+
+jobs:
+  coordinated-merge:
+    if: contains(github.event.pull_request.labels.*.name, 'coordinated-merge')
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.MERGE_TOKEN }}
+      NEON_REPO: exasim-project/NeoN
+      FOAM_REPO: exasim-project/FoamAdapter
```
greole: Depending on whether we already merged #146 
```suggestion
      FOAM_REPO: exasim-project/NeoFOAM
```

.github/workflows/coordinated-merge.yaml
```diff
@@ -0,0 +1,114 @@
+name: Coordinated Merge for FoamAdapter + NeoN
```
greole:
```suggestion
name: Coordinated Merge for NeoFOAM + NeoN
```

yhmtsai:
side note: if we have fixed version like #137 , we do not need the auto merge two repo branches triggered from one repo.

greole (commented):
LGTM! But please change FoamAdapter -> NeoFOAM

chihtaw:
> LGTM! But please change FoamAdapter -> NeoFOAM

Done

greole (approved):

PR #144, opened 2025-11-07
Use logging instead of cout and Info

greole:
This PR remove plain `std::cout ` and Info calls.

Additional changes:
- Add syncRuntime function.

### DIFF
--- a/examples/neoIcoFoam/neoIcoFoam.cpp
+++ b/examples/neoIcoFoam/neoIcoFoam.cpp
@@ -8,10 +8,15 @@
 #include "fvCFD.H"
 #include "pisoControl.H"
 
+#include <memory>
+
 using Foam::Info;
 using Foam::endl;
 using Foam::nl;
 
+// NOTE about namespace usage
+// here the namespaces are used deliberately verbose to
+// demonstrate where things are implemented
 namespace fvc = Foam::fvc;
 namespace dsl = NeoN::dsl;
 namespace fvcc = NeoN::finiteVolume::cellCentred;
@@ -21,7 +26,7 @@ namespace nf = NeoFOAM;
 
 int main(int argc, char* argv[])
 {
-    Kokkos::initialize(argc, argv);
+    NeoN::initialize(argc, argv);
     {
 #include "addCheckCaseOptions.H"
 #include "setRootCase.H"
@@ -39,9 +44,20 @@ int main(int argc, char* argv[])
         solverDict.subDict("p") = nf::mapFvSolution(solverDict.subDict("p"));
         solverDict.subDict("U") = nf::mapFvSolution(solverDict.subDict("U"));
 
-        Info << "creating nf pressure field" << endl;
         fvcc::VectorCollection& vectorCollection =
             fvcc::VectorCollection::instance(rt.db, "VectorCollection");
+
+        // create a new non-default logger with a given name, debug level and target
+        auto logger = std::make_shared<NeoN::Logging::Logger>(
+            "Debug Logger",             // name of the logger displayed in log
+            NeoN::Logging::Level::Info, // verbosity, Trace - most verbose
+            NeoN::Logging::Target::File // Console or File for JSON output
+        );
+        // add a logger to specific class that supports logging
+        setLogger(vectorCollection, logger);
+        // to get a more detailed picture of what is happening
+        // attach a logger to an executor
+        // setLogger(rt.exec, logger);
 
         fvcc::VolumeField<NeoN::scalar>& p =
             vectorCollection.registerVector<fvcc::VolumeField<NeoN::scalar>>(
@@ -53,7 +69,6 @@ int main(int argc, char* argv[])
                 }
             );
 
-        Info << "creating nf velocity field" << endl;
         fvcc::VolumeField<NeoN::Vec3>& U =
             vectorCollection.registerVector<fvcc::VolumeField<NeoN::Vec3>>(
                 nf::CreateFromFoamField<Foam::volVectorField> {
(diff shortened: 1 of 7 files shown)


examples/neoIcoFoam/neoIcoFoam.cpp
```diff
@@ -64,29 +79,28 @@ int main(int argc, char* argv[])
...
             );
 
-        Info << "creating nf nu field" << endl;
         auto nuBCs = fvcc::createCalculatedBCs<fvcc::SurfaceBoundary<NeoN::scalar>>(rt.nfMesh);
         fvcc::SurfaceField<NeoN::scalar> nu(rt.exec, "nu", rt.nfMesh, nuBCs);
         NeoN::fill(nu.internalVector(), viscosity.value());
         NeoN::fill(nu.boundaryData().value(), viscosity.value());
 
-        Info << "creating nf phi field" << endl;
+        NeoN::Logging::info("creating nf phi field");
```
chihtaw: `nf` is abbreviation of namespace used in the code.
For logging, it would be better to make it clear.

```suggestion
        NeoN::Logging::info("creating NeoFOAM phi field");
```

include/NeoFOAM/auxiliary/setup.hpp
```diff
@@ -15,6 +15,9 @@ namespace NeoFOAM
 /*@brief based on the Courant number this function synchronizes the deltaT value in both runtimes*/
 void setDeltaT(Foam::Time& ofRunTime, RunTime& nfRunTime, Foam::scalar coNum);
 
+/*@brief */
```
chihtaw: Please add a description to the function.
The description should mention how it is different from another function `setDeltaT`.

src/auxiliary/setup.cpp
```diff
@@ -50,20 +58,17 @@ std::unique_ptr<Foam::fvMesh> createMesh(const Foam::Time& runTime)
  */
 NeoN::Executor createExecutor(const Foam::word& execName)
 {
-    Foam::Info << "Creating Executor: " << execName << Foam::endl;
+    NeoN::Logging::info("Creating Executor"); //: {} ", execName);
```
chihtaw: Should `execName` be part of the output?
I think we need to print out which executor is used here.

src/auxiliary/setup.cpp
```diff
@@ -32,9 +40,9 @@ std::unique_ptr<MeshAdapter> createMesh(const NeoN::Executor& exec, const Foam::
 std::unique_ptr<Foam::fvMesh> createMesh(const Foam::Time& runTime)
 {
     std::unique_ptr<Foam::fvMesh> meshPtr;
-    Foam::Info << "Create mesh";
     Foam::word regionName(Foam::polyMesh::defaultRegion);
-    Foam::Info << " for time = " << runTime.timeName() << Foam::nl;
+    // NeoN::Logging::info("Create mesh for time = {}", runTime.timeName());
+    NeoN::Logging::info("Create mesh"); //, runTime.timeName());
```
chihtaw: Should `timeName` part of the output?

examples/neoIcoFoam/neoIcoFoam.cpp
```diff
@@ -39,10 +44,21 @@ int main(int argc, char* argv[])
...
+        // auto logger = std::make_shared<NeoN::Logging::Logger>(
+        //     "Debug Logger",             // name of the logger displayed in log
+        //     NeoN::Logging::Level::Info, // verbosity, Trace - most verbose
+        //     NeoN::Logging::Target::File // Console or File for JSON output
+        // );
+        // add a logger to specific class that supports logging
+        // setLogger(vectorCollection, logger);
+        // to get a more detailed picture of what is happening
+        // attach a logger to an executor
+        // setLogger(rt.exec, logger);
```
chihtaw: Please add a note which explains why we should keep commented code

src/compatibility/fvSolution.cpp
```diff
@@ -31,8 +34,7 @@ void updateSolver(NeoN::Dictionary& solverDict)
     auto it = solverMap.find(solverName);
     if (it != solverMap.end())
     {
-        std::cout << __FILE__ << ":\n\treplacing solver " << solverName << " by "
-                  << it->second.second << "\n";
+        NeoN::Logging::warn("Replacing solver {} by {}", solverName, it->second.second);
```
chihtaw: I know `it` is not part of the changes made in the PR.
Please use a more meaningful name if possible.

src/compatibility/fvSolution.cpp
```diff
@@ -154,10 +158,9 @@ NeoN::Dictionary mapFvSolution(const NeoN::Dictionary& solverDict)
     NeoN::Dictionary modSolverDict = solverDict;
 
     if (solverDict.contains("configFile")) return solverDict;
-    std::cout << __FILE__ << ":\n\tMapping OpenFOAM solver settings to NeoN settings\n"
-              << "\tCurrently, it is advisable to specify configFile for fine grained Ginkgo "
-                 "solver support\n";
-
+    NeoN::Logging::warn("Mapping OpenFOAM solver settings to NeoN settings.");
+    NeoN::Logging::warn("Currently, it is advisable to specify a configFile");
+    NeoN::Logging::warn("for fine grained Ginkgo solver control");
```
chihtaw: Can we make a single string with new line character in between, just like `std::cout`?
Especially for the second sentence "Currently, ...", I don't see a point to split it into two parts with two function calls.

include/NeoFOAM/auxiliary/setup.hpp
chihtaw:
```suggestion
} // namespace NeoFoam
```

src/auxiliary/setup.cpp
chihtaw, line 75: Should we replace it with the new logging system?

src/auxiliary/setup.cpp
greole: Yes, I am currently adding it again. The issue is that `Foam::word` are not working at the moment. We'll have to find a way to deal with it.

examples/neoIcoFoam/neoIcoFoam.cpp
greole: I think i'll move it to a separate branch!

src/compatibility/fvSolution.cpp
greole: what exactly? the second.second part?

src/auxiliary/setup.cpp
greole: I would deal with it later. The issue is that we don't have an equivalent to Error just yet.

src/compatibility/fvSolution.cpp
greole: ah ok i see understand replacing `it` with `mapEntry`

chihtaw (approved):
LGTM!

chihtaw:
If you want to try `coordinated-merge`, you need to make sure that both PRs pass all required CI checks.

PR #147, opened 2025-11-08
Python specific CI/CD setup

HenningScheufler:
Adds CI for:

* code coverage
* linting
* formating
* type checking

Actions:

- [ ] upload coverage.xml to code coverage

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -176,4 +176,5 @@ jobs:
         
     - name: Check types with ty
       run: |
+        uv run pip install .[all]
         uv run ty check
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -177,4 +177,4 @@ jobs:
     - name: Check types with ty
       run: |
         uv run pip install .[all]
-        uv run ty check
+        poe type_check
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -140,7 +140,7 @@ jobs:
       #     token: ${{ secrets.CODECOV_TOKEN }}
 
   lint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - name: Checkout
       uses: actions/checkout@v5
@@ -159,7 +159,7 @@ jobs:
         args: 'format --check'
 
   typing:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - name: Checkout
       uses: actions/checkout@v5
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -107,3 +107,37 @@ jobs:
       - name: Test with pytest
         run: |
           uv run pytest
+
+python-coverage:
+  runs-on: ubuntu-24.04 # may use an older version to test
+
+  name: Test coverage with OpenFOAM 2506 and Python 3.9
+  steps:
+    - uses: actions/checkout@v5
+    
+    - name: Set up OpenFOAM
+      uses: gerlero/setup-openfoam@v1
+      with:
+        openfoam-version: 2506
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.9
+
+    - name: Install test dependencies
+      run: |
+        uv run pip install -v .[all]
+      
+    - name: Test with pytest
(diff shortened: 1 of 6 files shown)


HenningScheufler:
@greole @chihtaw

.github/workflows/build.yaml
```diff
@@ -106,4 +106,75 @@ jobs:
...
+      with:
+        openfoam-version: 2506
+
+    - name: Lint with Ruff
+      uses: astral-sh/ruff-action@v3
+      
+    - name: Format with Ruff
+      uses: astral-sh/ruff-action@v3
+      with:
+        args: 'format --check'
```
greole, line 159: I think I would prefer a pre-commit based approach here, to have a single source of truth for local development and remote checking.  There is a hook for ruff too https://github.com/astral-sh/ruff-pre-commit

.github/workflows/build.yaml
```diff
@@ -106,4 +106,75 @@ jobs:
...
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+      with:
+        python-version: 3.9
+        
+    - name: Check types with ty
+      run: |
+        uv run pip install .[all]
+        uv run ty check
```
greole: any particular reason for `ty` over `mypy` for example? mypy has  pre-commit support and doesn't warn

> ty is in preview and is not ready for production use.
> 
> We're working hard to make ty stable and feature-complete, but until then, expect to encounter bugs, missing features, and fatal errors.

.github/workflows/build.yaml
```diff
@@ -106,4 +106,75 @@ jobs:
...
+
+  lint:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v5
+
+    - name: Set up OpenFOAM
+      uses: gerlero/setup-openfoam@v1
+      with:
```
greole, line 150: Why is OF needed for linting of our python stuff?

.github/workflows/build.yaml
```diff
@@ -106,4 +106,75 @@ jobs:
...
+  typing:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v5
+
+    - name: Set up OpenFOAM
+      uses: gerlero/setup-openfoam@v1
+      with:
+        openfoam-version: 2506
```
greole, line 170: Same here, is OF really needed for our python type checking.

.github/workflows/build.yaml
```diff
@@ -106,4 +106,75 @@ jobs:
...
+    runs-on: ubuntu-24.04 # may use an older version to test
+
+    name: Test coverage with OpenFOAM 2506 and Python 3.9
+    steps:
+      - uses: actions/checkout@v5
+      
+      - name: Set up OpenFOAM
+        uses: gerlero/setup-openfoam@v1
+        with:
+          openfoam-version: 2506
```
greole, line 121: needed?

greole (commented):
Some quick comments.

.github/workflows/build.yaml
HenningScheufler: some test require pybFoam which requires the installation of OpenFOAM

.github/workflows/build.yaml
HenningScheufler: Avoid import errors of OpenFOAM caused by pybFoam.

However, im unsure if it is an issue

.github/workflows/build.yaml
HenningScheufler: We should ensure that the format is consistent.

But the addition of a precommit would be excellent
Can you add it?

.github/workflows/build.yaml
HenningScheufler: pybFoam requires OpenFOAM and this is to avoid import errors.
However, im not sure if it is a mistake

.github/workflows/build.yaml
HenningScheufler: mypy could be an option

But we should try to enforce type checking as early as possible

HenningScheufler: switch to mypy.

switch the tool should be straightforward as ideally poe is used

poe type_check and what happens under the hood is a minor concern

HenningScheufler:
@greole 
Gregor can you approve the PR?

The missing actions will be added a new PR

- [ ] host coverage report (Gregor)
- [x] pre commit  (done #150 )

greole (approved):
LGTM! But some files seem to miss license information.

PR #152, opened 2025-11-11
rename

greole:
I managed to move the lrz repo url, so the trigger scripts should work again without hardcoding the url.


chihtaw (changes requested):
https://github.com/exasim-project/NeoFOAM/blob/eaac0e43d19704c40f172f6446f3d3e0bfe0589b/ci/github/trigger_pipeline.sh#L25

I think that this hardcoded line in the script of triggering pipelines should also be fixed.
Could you fix it?

chihtaw (approved):
LGTM!

PR #153, opened 2025-11-11
remove deprecated solver

greole:
This PR removes the scalarAdvection and heatTransfer solver. I think on the cpp side we should focus on polishing neoIcoFoam first and the continue with pimple/pisoFoam.


HendriceH (approved):
Makes sense to me.

PR #159, opened 2025-11-14
Fix the name of the fluid flow solver in tutorials

chihtaw:
A wrong name is used for the fluid flow solver: `icoNeoFoam`

Since there is no solver with this name, the tutorial cases using the wrong name will immediately fail.

Fix: Use `neoIcoFoam` to run the tutorial cases.

### DIFF
--- a/tutorials/cylinder2D/Allrun
+++ b/tutorials/cylinder2D/Allrun
@@ -7,7 +7,7 @@ restore0Dir
 
 runApplication blockMesh
 
-# runApplication ../../build/profiling/bin/icoNeoFoam
-runApplication icoFoam
+runApplication ../../build/profiling/bin/neoIcoFoam
+# runApplication icoFoam
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tutorials/cylinder2D/Allrun
```diff
@@ -7,7 +7,7 @@ restore0Dir
 
 runApplication blockMesh
 
-# runApplication ../../build/profiling/bin/icoNeoFoam
-runApplication icoFoam
+runApplication ../../build/profiling/bin/neoIcoFoam
```
dheerajraghunathan, line 10: Are the tutorials always expected to run with the `profiling` preset?

chihtaw: No. The tutorials can be run with either `develop` or `profiling` preset.
It would be better if we can have a more elegant to select a particular fluid flow solver built with a certain preset. But it is not the point of this PR.

dheerajraghunathan: Ok. Similar to OpenFOAM, we can also use variables like $FOAM_APPBIN in such cases. Maybe in a different PR.

Otherwise LGTM.

dheerajraghunathan (approved):

PR #166, opened 2025-11-18
Fix issue with int to scalar conversion

greole:
This PR fixes the issue with the failing badAnyCast. The fix proposed in https://github.com/exasim-project/NeoN/actions/runs/19477256232/job/55829317984?pr=403 slightly changes the use of dictionaries. Now dict.get<T> returns a copy (which allows to convert for example from int to scalar). Most cases we want to access a reference can be handled by dictionary.subDict which returns a reference.

### DIFF
--- a/test/test_advection.cpp
+++ b/test/test_advection.cpp
@@ -118,7 +118,7 @@ TEST_CASE("Advection Equation")
     SECTION("Scalar advection with " + execName + " and " + "forwardEuler")
     {
         std::string timeIntegration = "forwardEuler";
-        fvSchemesDict.get<NeoN::Dictionary>("ddtSchemes").insert("type", timeIntegration);
+        fvSchemesDict.subDict("ddtSchemes").insert("type", timeIntegration);
 
         while (runTime.run())
         {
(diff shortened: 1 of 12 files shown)


test/test_advection.cpp
```diff
@@ -49,6 +49,9 @@ void initFields(Foam::volScalarField& T, Foam::volVectorField& U, Foam::surfaceS
 TEST_CASE("Advection Equation")
 {
     Foam::Time& runTime = *timePtr;
+    Info << "Foo bbar bbaz\n";
+    // Info.level = 0;
+    // Info<< "Foo bbar bbaz 222 \n";
```
greole:
```suggestion
```

tutorials/cylinder2D/system/controlDict
```diff
@@ -16,7 +16,7 @@ FoamFile
 
 application     icoFoam;
 
-executor        GPU;
+executor        Serial;
```
greole, line 19:
```suggestion
executor        GPU;
```

PR #167, opened 2025-11-19
Enhance coordinated merge on NeoFOAM and NeoN

chihtaw:
The current version of coordinated merge only checks if the two PRs of interest get approved before merging.
But sometimes a PR can get approved before it passes all CI checks. 

It means that the two PRs will be merged together automatically via coordinated merge even if some CI checks fail.
The consequence is that the merged PRs might not be actually finished and can break the code.

To avoid such situation, this PR enhance the GitHub workflow of coordinated merge by adding two checks which confirm that the two PRs of interest have passed all the required CI checks before merging.

In addition, the GitHub workflow is refactored to be more concise and more readable.


greole (approved):
LGTM!

PR #176, opened 2025-11-24
Make some static checks to be skipped on demand

chihtaw:
With the current CI setup, the static checks cannot be skipped when the `Skip-build` label is added after a PR is opened.

This PR makes the static checks to be skipped whenever the `Skip-build` is present.


greole (approved):
LGTM!

PR #181, opened 2025-12-01
Automate validation of cavity flow tutorial

### LINKED ISSUE #162: Need automatic validation of fluid flow solvers
There is a need of automatic validation process for fluid flow solvers in NeoFOAM.

It should be very useful for development work afterwards.

dheerajraghunathan:
The PR is to automate the validation process of `neoIcoFoam` solver using the cavity flow tutorial. The python script `cleanRunPlot.py` is executable and perform all three functions - clean the case directory, run `blockmesh` + `neoIcoFoam`  and finally, plot centreline velocities and validate them against the reference (Ghia et al.). The reference data `ghia_u.txt`  and `ghia_v.txt` contain the data for all the Reynolds number reported in the literature. The  `cleanRunPlot.py` script reads `nu` value and computes `Re`; also detects the `endTime`.

Individual functionalities are also permitted.
```
- ./cleanRunPlot.py --clean
- ./cleanRunPlot.py --run
- ./cleanRunPlot.py --plot
```
Note: Packages `matplotlib` and `foamlib` are required.

Close #162


chihtaw (commented):
Since the script includes three different functionalities, I think it would be better to split the long script into three shorter scripts, each of which focuses on a single functionality. 

Shorter scripts of single functionality should be easier to maintain than a single long script of multiple functionalities.

The script name `cleanRunPlot.py` is a little confusing. If we split it into three shorter scripts, then we can simply name each script by its functionality or use a more descriptive name, e.g., `cleanTutorial.py` for the script of cleaning tutorial cases.

greole:
Some of the features implemented here are already available in OBR we could also reuse it. We can have a chat in our next meeting.

greole (approved):

greole:
Is there a currently a plan to integrate that into a workflow?

dheerajraghunathan:
Currently, this is a standalone facility. We need to discuss the integration into a workflow in detail.

dheerajraghunathan:
> Since the script includes three different functionalities, I think it would be better to split the long script into three shorter scripts, each of which focuses on a single functionality.
> 
> Shorter scripts of single functionality should be easier to maintain than a single long script of multiple functionalities.
> 
> The script name `cleanRunPlot.py` is a little confusing. If we split it into three shorter scripts, then we can simply name each script by its functionality or use a more descriptive name, e.g., `cleanTutorial.py` for the script of cleaning tutorial cases.

I would like to consider splitting the code and improving this on a separate PR intended for workflow integration.

PR #183, opened 2025-12-02
deactivate test in z direction for now

greole:
Currently, the pressure Advection test case is not reliable. Esp. in the z direction we deviate from OF since we don't explicitly zero out values. Additionally, we solve coupled vs segregated and have a different stopping criterion, thus we should have a proper margin.


chihtaw (approved):
Please add a description of why the test in z direction needs to be deactivated.

PR #187, opened 2025-12-07
Refactor Cmake import function for OF imports

greole:
If we add  further functionality from OF we will need to include and link more parts of it. This will become increasingly repetitive and prone to errors. Thus this PR adds a cmake function to simplify imports.

```
importoflibrary(finiteVolume)
importoflibrary(incompressibleTransportModels INCLUDE transportModels/incompressible)
importoflibrary(Pstream INCLUDE Pstream/mpi LIBPATH $ENV{FOAM_LIBBIN}/$ENV{FOAM_MPI})
```


HenningScheufler (approved):
LGTM

PR #194, opened 2025-12-18
Fix fixedValue boundary reader

greole:
Currently reading `fixedValue` boundaries might print an error but then proceeding.


HendriceH (approved):
LGTM!

PR #196, opened 2025-12-18
Cover benchmark code in default CI pipelines

chihtaw:
In the current CI setup, we do not build benchmarks for CI pipelines by default. It means that the default CI pipelines are not able to capture the failure of benchmark code.

This PR makes the default CI pipelines to build benchmark code additionally.


greole (approved):
LGTM! Do we also need to turn of the benchmarks when running the tests?

chihtaw:
> LGTM! Do we also need to turn of the benchmarks when running the tests?

No. Although benchmarks are built, they are not part of running tests.

chihtaw:
> > LGTM! Do we also need to turn of the benchmarks when running the tests?
> 
> No. Although benchmarks are built, they are not part of running tests.

I have to correct myself. We need to exclude benchmarks when using ctest to run unit tests.

PR #198, opened 2025-12-19
Fix(benchmarks): resolve build failure of benchmark binaries

dheerajraghunathan:
The benchmark binaries currently fail to build. This PR is a quick fix for this issue.


greole (approved):
LGTM!

PR #200, opened 2025-12-28
Remove KOKKOS_LAMBDA in favour of NEON_LAMBDA

greole:
NeoFOAM should not depend on NeoNs implementation details, thus this PR removes `KOKKOS_LAMBDA` calls in favour of `NEON_LAMBDA`.


HenningScheufler (approved):
LGTM

PR #202, opened 2026-01-09
Reduce memory footprint on GPU

greole:
To reduce the memory footprint on GPU we keep some data on the CPU.

### DIFF
--- a/src/datastructures/meshAdapter.cpp
+++ b/src/datastructures/meshAdapter.cpp
@@ -56,13 +56,17 @@ int32_t computeNBoundaryFaces(const Foam::fvMesh& mesh)
     return nBoundaryFaces;
 }
 
-NeoN::UnstructuredMesh readOpenFOAMMesh(const NeoN::Executor exec, const Foam::fvMesh& mesh)
+NeoN::UnstructuredMesh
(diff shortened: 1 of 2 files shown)


src/datastructures/meshAdapter.cpp
```diff
@@ -116,7 +120,7 @@ NeoN::UnstructuredMesh readOpenFOAMMesh(const NeoN::Executor exec, const Foam::f
     );
 
     NeoN::UnstructuredMesh uMesh(
-        fromFoamField(exec, mesh.points()),
+        fromFoamField(optExec, mesh.points()),
         fromFoamField(exec, mesh.cellVolumes()),
         fromFoamField(exec, mesh.cellCentres()),
```
HenningScheufler, line 125: Is cell centers needed?

HenningScheufler (approved):

src/datastructures/meshAdapter.cpp
greole: yes for the weights. but we reset that after computing the weights on the neon site.

greole: at some point we should introduce a new hostViewVector class that keeps the pointer to original data.

PR #205, opened 2026-01-17
fixed the test and added kernels

HenningScheufler:
works with the NeoN branch fix/benchmarks

* fixes the benchmarks missing Neon::finalize
* adds allocation benchmarks showing a allocation performance increase of a factor 20 on CPUs
* adds experimental kernels with fused operators


benchmarks/catch_main.hpp
```diff
@@ -77,5 +77,6 @@ int main(int argc, char* argv[])
         // Run benchmarks if there are any
         result = session.run();
     }
+    NeoN::finalize();
```
greole, line 80: can you move that to a separate PR?

PR #210, opened 2026-01-23
cmake fetches version from submodule and update documentation in proper order

yhmtsai:
This PR enable cmake will try to grab the version automatically from submodule if the project is from github not from zip and git is installed in the environment.
If they can not grab the version, it will fallback to the predefined version - main.

Question: should it be main branch from NeoN?

Additionally, I update the documentation in the proper order. The above feature only change the version selection not the order

`NEOFOAM_NEON_DIR` -> `NEOFOAM_NEON_VERSION` -> src/NeoN contains repo (from submodule initialization or not) -> version from submodule SHA or predefined version

After this pr, it mainly changes the user setup without setting branch name. 
Current CI or user needs to specify to develop in the original behavior.

### DIFF
--- a/cmake/CxxThirdParty.cmake
+++ b/cmake/CxxThirdParty.cmake
@@ -9,9 +9,27 @@ endif()
 
 if(NEOFOAM_NEON_VIA_CPM)
   if(NOT DEFINED NEOFOAM_NEON_VERSION)
-    set(NEOFOAM_NEON_VERSION
-        "main"
-        CACHE INTERNAL "")
+    # grab the SHA from submodule automatically
+    find_package(Git QUIET)
+    if(EXISTS "${NeoFOAM_SOURCE_DIR}/.git" AND GIT_FOUND)
+      # Note: it will not initialize the submodule and does not require initialization
+      execute_process(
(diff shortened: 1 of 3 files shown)


greole (approved):

cmake/CxxThirdParty.cmake
```diff
@@ -9,9 +9,27 @@ endif()
...
+      string(SUBSTRING "${GIT_SUBMODULE_STRING}" 0 1 SUBMODULE_PREFIX)
+      string(SUBSTRING "${GIT_SUBMODULE_STRING}" 1 40 SUBMODULE_HASH)
+      set(NEOFOAM_NEON_VERSION
+          "${SUBMODULE_HASH}"
+          CACHE INTERNAL "")
+    else()
+      # if is is not from git or user does not have git, it fallback to predefined value.
+      set(NEOFOAM_NEON_VERSION
+          "main"
+          CACHE INTERNAL "")
```
yhmtsai, line 31: @greole should it still be `main`?

yhmtsai: I will keep as what it is like original now.

PR #212, opened 2026-01-26
CMake execution and cleanup of benchmarks

greole:
This PR cleans up the NeoFOAM benchmarks, it includes:

- Use helper functions to create `ofPhi`, `ofGamma`, `ofT` ...
- single script to generate, execute, and postprocess benchmark parameter studies
-  benchmarks can be executed via our standard ctest procedure
- add `constFromMany` to construct nf fields from a set of OF fields
- add benchmark for pressure velocity coupling components


.github/workflows/build.yaml
```diff
@@ -86,6 +86,9 @@ jobs:
           ninja --version
           cmake --version
 
+      - name: Install python dependencies
+        run:  pip install foamlib pandas foamlib pydantic defusedxml
```
greole:
```suggestion
        run:  pip install pandas foamlib pydantic defusedxml
```

chihtaw:
A PR label `skip-nvidia` is added to the PR, but there is no implementation for that.
I can implement that in another PR.

.github/workflows/build.yaml
```diff
@@ -86,6 +86,9 @@ jobs:
           ninja --version
           cmake --version
 
+      - name: Install python dependencies
+        run:  pip install foamlib pandas pydantic defusedxml
```
chihtaw, line 90: Why are these python packages installed here?

greole: We need them to generate the cases for benchmarking.

benchmarks/bench_pressureVelocityCoupling.cpp
```diff
@@ -0,0 +1,114 @@
...
+        {
+            BENCHMARK("CPU")
+            {
+                Foam::constrainHbyA(forAU * ofUEqn.H(), ofU, ofP);
+                return;
+            };
+        };
+    }
+
+    SECTION("NeoN")
```
chihtaw, line 65:
```suggestion
    SECTION("NeoFOAM")
```
It should be `NeoFOAM`?

.github/workflows/build.yaml
greole: Once we have [this](https://github.com/exasim-project/NeoFOAM/pull/133) we'll need someway to handle python dependencies anyways.

benchmarks/bench_pressureVelocityCoupling.cpp
greole: on the other hand we currently call all sections NeoN

.github/workflows/build.yaml
chihtaw: This GitHub workflow is used to build NeoFOAM with develop preset and run unit test.
Will this GitHub workflow generate cases for benchmarking?

greole: Since it uses `-DNEOFOAM_BUILD_BENCHMARKS=On`  to build the benchmark binaries it also needs to be able to generate the cases. I think it is sensible behaviour that if you want to build with benchmarks you need to be also able to run the benchmarks.

chihtaw:
I found some inconsistency and redundancy which exist before the changes introduced by this PR.
I did some refactoring for that in another PR. #215 
Can you include those changes in this PR?

.github/workflows/build.yaml
chihtaw: I see, but the GitHub workflow is used to verify the correctness of code. We build benchmark code in this workflow only for completeness, but we don't run benchmarks here. It looks like we install those python packages here, but will never use them.

chihtaw:
The PR #216 of skipping CI on Intel GPU is done.

.gitlab-ci.yml
```diff
@@ -79,13 +79,16 @@
...
     RUN_IDENTIFIER: "${CI_PIPELINE_ID}"
+    PR_NUMBER: "$PR_NUMBER"
     REPO_NAME: "NeoFOAM-BenchmarkData"
     TARGET_REPO: "github.com/exasim-project/${REPO_NAME}.git"
     TARGET_BRANCH: "ci-benchmarks"
-    TARGET_RESULTS_DIR: "NeoFOAM/${RUN_IDENTIFIER}/${CI_RUNNER_DESCRIPTION}"
+    TARGET_RESULTS_DIR: "NeoFOAM/${PR_NUMBER}/${CI_RUNNER_DESCRIPTION}"
 
+<<<<<<< HEAD
+<<<<<<< HEAD
```
chihtaw: I think these two lines should not be here. It looks like something left after handling merge conflicts.

greole: yes correct! i removed it already.

benchmarks/benchmarkSuite/templates/2DSquare/system/fvSchemes
```diff
@@ -17,7 +17,7 @@ FoamFile
 ddtSchemes
 {
     default         Euler;
-    type            backwardEuler;
+    ddt(U)          backward;
```
chihtaw, line 20: Can NeoFOAM identify `backward` for ddt(U)?
I thought only `BDF1` and `BDF2` are usable for ddt(U).

chihtaw (approved):
LGTM!

Before you merge it, can you make CI run benchmarks at least on one GPU so that we can see what we can get from this PR?

benchmarks/benchmarkSuite/templates/2DSquare/system/fvSchemes
greole: Yes we have a mapping between backward which maps to BDF.

chihtaw: So we have mappings for ddt(U) below. 
`backward` -> `BDF1`
`BDF1` -> `BDF1`
`BDF2` -> `BDF2`

Is it correct?

greole: We use the mapping below. So it should be rather Euler for ddt(U)
```
        {"Euler", "BDF1"},
        {"backward", "BDF2"},
```

PR #214, opened 2026-01-28
Enable developers to skip CI on NVIDIA GPU

chihtaw:
When NVIDIA GPUs are not available for CI for quite a period of time, developers might want to skip CI on NVIDIA GPUs.
This PR enables skipping CI on NVIDIA GPUs by adding the label `skip-nvidia` to the PR.


greole:
does it work for AMD and Intel?

greole (approved):

chihtaw:
> does it work for AMD and Intel?

When the PR label `skip-nvidia` is added, only CI pipelines on AMD and Intel GPUs will be triggered.

greole:
i mean can we also add an `skip-intel` and `skip-amd` option.

PR #218, opened 2026-01-29
Refactor GitLab CI configurations

chihtaw:
This PR modularizes the GitLab CI config file. It redefines config sections and reuses them to compose different CI jobs.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,35 +2,46 @@
 #
 # SPDX-License-Identifier: Unlicense
 
-# Available GPUs and corresponding Docker images
-.use-nvidia-gpu-and-image:
+# ============================================================
+#  GPU-specific images and runner configuration
+# ============================================================
+
+.gpu-base:
+  variables:
+    PYTHONPATH: /root/.local/:/root/.local/lib/:/root/.local/lib/python3.12:/root/.local/lib/python3.12/site-packages
+    WM_PROJECT_VERSION: v2412
+
+.gpu-nvidia:
+  extends: .gpu-base
   tags: ["nvidia-gpus"]
   image: chihtaw/cuda-openfoam-ginkgo:12.8.1-v2412-arch_89
   variables:
-    PYTHONPATH: /root/.local/:/root/.local/lib/:/root/.local/lib/python3.12:/root/.local/lib/python3.12/site-packages
     FOAM_INST_DIR: /usr/local/app/openfoam
-    WM_PROJECT_VERSION: v2412
+    GPU_VENDOR: nvidia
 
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -2,35 +2,43 @@
...
-.use-intel-gpu-and-image:
+.gpu-intel:
+  extends: .gpu-base
   tags: ["intel-gpus"]
   image: dheerajraghunathan/oneapi-openfoam-ginkgo:2025.3-v2412-arch_pvc
-  before_script:
-    - source /opt/intel/oneapi/setvars.sh --force
-    - eval "$(bash -c 'source /opt/openfoam/OpenFOAM-v2412/etc/bashrc >/dev/null 2>&1; export -p')"
   variables:
-    PYTHONPATH: /opt/intel/oneapi/advisor/2025.3/pythonapi
```
dheerajraghunathan: Is the Python path for Intel specified elsewhere?
Please change to `/opt/intel/oneapi/advisor/2025.4/pythonapi` if this is needed!

.gitlab-ci.yml
```diff
@@ -2,35 +2,43 @@
...
     FOAM_INST_DIR: /opt/openfoam
-    WM_PROJECT_VERSION: v2412
+    GPU_VENDOR: intel
+
+# ============================================================
+#  Common OpenFOAM environment setup
+# ============================================================
 
-# Common OpenFOAM environment variables
 .openfoam-env-setup:
```
dheerajraghunathan: Is this applicable to Intel also? We use icx compiler instead of gcc.

.gitlab-ci.yml
chihtaw: Yes. The `before_script` section of intel GPU is merged with the other `before_script` section of OpenFOAM environment setup so that only a single `before_script` section exists in the config file. By doing so, we can avoid unintentionally overwrite of a `before_script` section by another one.

.gitlab-ci.yml
chihtaw: Yes. CI passes for all GPUs.

.gitlab-ci.yml
dheerajraghunathan: I think we need this for the intel image.
PYTHONPATH: /opt/intel/oneapi/advisor/2025.4/pythonapi

greole: it probably also needs the /root/.local files to find any pip installed stuff.

greole (approved):
LGTM! But needs to be rebased on current develop first

chihtaw:
> LGTM! But needs to be rebased on current develop first

Sure!

PR #236, opened 2026-02-10
Read, write and validate pydantic models

HenningScheufler:
NeoFOAM simulations rely on structured data/dictionaries to define solver settings, boundary conditions, and physical models. Manually parsing, validating, and writing these files is repetitive and error-prone.

The configuration data of the solver is defined by pydantic with a mulitple IOStrategy to allow to read/write different formats. pydantic naturally supports validation and the errors can be gathered to validation the inputs file and give feedback to the user

### Usage 

```python

   from pydantic import Field
   from neofoam.io import BaseConfig, YAML, IOStrategy

   @IOStrategy(YAML("solver_config.yaml"))
   class SolverConfig(BaseConfig):
       tolerance: float = Field(gt=0)
       maxIterations: int = Field(gt=0, le=10000)
       solver: str

   # Load, modify, save
   cfg = SolverConfig.load(case_dir="path/to/case")
   cfg.maxIterations = 500
   cfg.save(case_dir="path/to/case")
```

```python

   cfg = SolverConfig.load(case_dir="case", validate=False)
   errors = cfg.check_validation()  # returns list[ValidationErrors]
```


src/neofoam/io/strategies/json_strategy.py
```diff
@@ -0,0 +1,79 @@
...
+
+import json
+from typing import Any, Optional
+from pathlib import Path
+
+from neofoam.io.strategies.subdict import SubdictMixin
+
+
+class JSONStrategy(SubdictMixin):
+    """JSON reading and writing strategy.
```
greole: Can you specify what is read? A JSON dictionary from file?

doc/development/config_file.rst
```diff
@@ -0,0 +1,215 @@
+Configuration File IO
+=====================
+
+NeoFOAM simulations rely on structured data/dictionaries to define solver settings, boundary conditions, and physical models.
+Manually parsing, validating, and writing these files is repetitive and error-prone.
+
+The ``neofoam.io`` module builds on `Pydantic <https://docs.pydantic.dev/>`_ to solve this.
+Pydantic was chosen because it provides:
+
+* **Declarative field definitions** — types and constraints (e.g. ``gt=0``, ``le=100``) are specified inline, not in separate validation logic.
```
greole:
```suggestion
* **Declarative field definitions** — types and constraints (e.g. ``gt=0``, ``le=100``) are specified inline, not in a separate validation logic.
```

doc/development/config_file.rst
```diff
@@ -0,0 +1,215 @@
...
+.. code-block:: python
+
+   errors = SolverConfig.collect_errors(case_dir="case")
+
+
+IO Strategies
+-------------
+
+The ``@IOStrategy`` decorator accepts a strategy configuration.
+Each strategy implements the ``ReadingStrategy`` and ``WritingStrategy`` protocols.
```
greole: This is a bit unclear to me. -> Can you give an example of the strategies for which the protocols are implemented? Also, why are the ReadingStrategy and WritingStrategy called protocols?

doc/development/config_file.rst
```diff
@@ -0,0 +1,215 @@
...
+     - YAML
+     - ``YAML("fvSolution.yaml", subdict="PIMPLE")``
+   * - ``JSON(file)``
+     - JSON
+     - ``JSON("config.json")``
+   * - ``JSON(file, subdict=...)``
+     - JSON
+     - ``JSON("config.json", subdict="services.db")``
+
+Both strategies support three subdict modes:
```
greole: strategies or protocols?

test/io/test_io_decorator_nested.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Reload from the new file and compare
+    reloaded_meta = metadata_class.load(case_dir=tmp_path, file=filename)
+    reloaded_s1 = stage1_class.load(case_dir=tmp_path, file=filename)
+
+    assert reloaded_meta.model_dump() == original_meta.model_dump()
+    assert reloaded_s1.model_dump() == original_s1.model_dump()
+
+    # Clean up the written file
+    (tmp_path / filename).unlink()
```
greole, line 140: should this be part of the test it self or rather handled via a fixture?

test/io/test_io_decorator_shared_file.py
```diff
@@ -0,0 +1,179 @@
...
+    service_a = service_a_class.load(case_dir=test_dir)
+    service_a.timeout = 60
+    service_a.save(case_dir=test_dir)
+
+    # Verify service_a changed but other sections unchanged
+    updated_a = service_a_class.load(case_dir=test_dir)
+    updated_b = service_b_class.load(case_dir=test_dir)
+    updated_c = service_c_class.load(case_dir=test_dir)
+
+    assert updated_a.timeout == 60  # Changed
```
greole: Is the comment correct? To me it looks like it is unchanged?  Also it is probably to use a variable to make sure the unchanged values are the still the same.

```suggestion
    assert updated_a.timeout == timeout  # Unchanged
```

greole (commented):
just some minor comments

greole (approved):

doc/development/config_file.rst
HenningScheufler: protocols are an implementation of an interface to use duck typing

changed to interface

test/io/test_io_decorator_nested.py
HenningScheufler: The test already loads the files and the fixture only change defines the tmp path

so something like
def tmp_path():

yield

filename.unlink() # is not possible

doc/development/config_file.rst
greole: I just wanted to point out that the naming of things and what they represent should be consistent. If you say "the protocol `ReadingStrategy`" I ask myself, why is it not called  `ReadingProtocol`, or you refer to `ReadingStrategy` as strategy.

PR #237, opened 2026-02-11
Fix/solver settings

HendriceH:
Mapping relTol to "relative_residual_norm" is wrong the correct behaviour is recovered by "initial_residual_norm".
Also when using BDF2, the startup phase where BDF1 is used is detected via the number of old time rotations. Thus the fields mustn't be rotated on registration.


greole (approved):
LGTM!

PR #244, opened 2026-02-19
Enable Kokkos Tools for profiling and debugging

chihtaw:
**Add Kokkos Tools integration for debugging and profiling**
Kokkos Tools is enabled when the CMake preset `develop` or `profiling` is used.

Setup of Kokkos Tools is handled by the script `run_with_kokkos_tool.sh`.

**Usage**:
`./path-to-the-script <debug/profile> <Kokkos-tool-name> ./path-to-the-fluid-flow-solver`

Available debugging tools:
kernel logger

Available profiling tools:
- simple kernel timer
- space time stack
- memory high water mark


cmake/Profiling.cmake
```diff
@@ -0,0 +1,73 @@
+# SPDX-License-Identifier: Unlicense
+# SPDX-FileCopyrightText: 2026 NeoFOAM authors
+#
+# NeoFOAM Profiling Integration via Kokkos Tools
+#
+include_guard(GLOBAL)
+
+option(NEOFOAM_ENABLE_PROFILING "Enable profiling via Kokkos Tools" OFF)
```
greole: can you move this to the main CMakeLists.txt. I think it is easier to discover all the options without having to run (c)cmake.

cmake/KokkosTools.cmake
```diff
@@ -0,0 +1,73 @@
...
+  BINARY_DIR ${KOKKOS_TOOLS_BUILD_DIR}
+  CMAKE_ARGS -DBUILD_SHARED_LIBS=ON
+             -DKokkosTools_ENABLE_MPI=OFF
+             -DKokkosTools_ENABLE_PAPI=OFF
+             -DKokkosTools_ENABLE_CALIPER=OFF
+             -DKokkosTools_ENABLE_APEX=OFF
+             -DKokkosTools_ENABLE_EXAMPLES=OFF
+             -DKokkosTools_ENABLE_TESTS=OFF
+  INSTALL_COMMAND ""
+  USES_TERMINAL_BUILD TRUE)
```
greole, line 54: use CPM for that. See `cmake/CxxThirdParty.cmake` so that we have a consistent way.

cmake/Profiling.cmake
greole: Ok, it seems that this one is actually redundant.

greole (commented):
some quick comments about consistent TPL handling.

cmake/KokkosTools.cmake
chihtaw: Kokkos-tools is not a TPL for NeoFOAM because NeoFOAM does not depend on kokkos-tools. 
Kokkos-tools is set up as an external project which is built independent of NeoFOAM and loaded at runtime for profiling.
CPM is not suitable for this scenario.

greole: OK, but i think it makes sense to keep everything that downloads external stuff in a single place.

chihtaw: I am not sure it would be better if we do so. 

Currently, we have a TPL cmake file for downloading TPLs and a profiling cmake file for downloading profiling tools. This separation is good for managing downloads based on different concerns.

If we simply keep all of them in a single place, it would not be easier for navigation based on a certain concern, e.g., check which profiling tool will be downloaded.

Another way to keep commands that download external stuff can be that we add another layer on top of both the TPL cmake file and the profiling cmake file and use it to specify all downloads.

chihtaw:
Add support for develop build for debugging

cmake/KokkosTools.cmake
```diff
@@ -5,11 +5,11 @@
 #
 include_guard(GLOBAL)
 
-if(NOT NEOFOAM_ENABLE_PROFILING)
+if(NOT KOKKOS_TOOLS_ENABLE)
```
greole: Please Prefix with `NEOFOAM` to avoid any potential conflicts.

CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif()
 option(NEOFOAM_BUILD_EXAMPLES "Build the NeoN examples" ON)
 option(NEOFOAM_BUILD_TESTS "Build the unit tests" OFF)
 option(NEOFOAM_BUILD_BENCHMARKS "Build benchmarks" OFF)
-option(KOKKOS_TOOLS_ENABLE "Enable Kokkos Tools" OFF)
+option(NEOFOAM_KOKKOS_TOOLS_ENABLE "Enable Kokkos Tools" OFF)
```
greole: Or better `NEOFOAM_ENABLE_KOKKOS_TOOLS`

cmake/KokkosTools.cmake
greole: > If we simply keep all of them in a single place, it would not be easier for navigation based on a certain concern, e.g., check which profiling tool will be downloaded.

It don't think it is really a concern at the moment since we only have one TPL, eg. NeoN. But i would prefer that we have a consistent way to deal with external stuff.

HenningScheufler:
Can you update the documenation with a section how to use it

greole (approved):

PR #252, opened 2026-03-02
Make the option of enabling kokkos tools easier to use and check

chihtaw:
- Rename the option according to the naming convention for configuring options
- Make the option of enabling kokkos tools visible in the configuration banner


greole (approved):
LGTM!

PR #253, opened 2026-03-02
Improve the use of kokkos tools

chihtaw:
**Motivation**
Most of the currently available kokkos tools write their output to the terminal. It is not always easy to navigate information in the terminal output when the output is very long. In addition, the output of the simple kernel timer is a specialized file format by default, which needs to be read by the provided reader.

**Approach**
This PR enhances the script of using kokkos tools by making a chosen kokkos tool to write output to a log file, whose filename can be specified as an optional command-line argument to the script. Once the output is stored in a file, it will be easier to navigate or post-process the debugging or profiling information.

**Documentation update**
- Add documentation of using Kokkos tools for NeoFOAM applications.
- Update the figure used in the continuous integration section


greole (approved):
LGTM!

PR #258, opened 2026-03-14
Fix fmt error on AMD

dheerajraghunathan:
This PR fixes a build issue on AMD caused by using hipcc as the global C++ compiler. That configuration caused dependencies like `fmt` to trigger errors during device compilation, as shown below.

`error: reference to __host__ function 'assert_fail' in __host__ __device__ function`.

Also, added `-DNEOFOAM_BUILD_BENCHMARKS=ON` for the build and test step. This was missing for AMD GPUs.


greole (approved):
LGTM!

PR #275, opened 2026-04-26
Test diagonal and upper triangular matrix coefficients in pressureVelocityCoupling test

greole:
This PR is the counterpart to https://github.com/exasim-project/NeoN/pull/495 and adds a test to the pressureVelocityCoupling tests to ensure the correctness of the diagonal and upper triangular matrix coefficients.

Additionally: 
- fix order of arguments for the explicit grad benchmark

### DIFF
--- a/test/test_pressureVelocityCoupling.cpp
+++ b/test/test_pressureVelocityCoupling.cpp
@@ -95,6 +95,9 @@ TEST_CASE("PressureVelocityCoupling")
             ApproxVector(1e-15)
         );
 
+        // NeoN stores boundary diagonal contributions directly in the matrix, whereas
+        // OpenFOAM keeps them separate. Remove them before comparing against OpenFOAM
+        // coefficients.
         nf::compare(
             NeoN::la::removeBoundaryContributions(nfUEqn.linearSystem()).matrix().diag(),
             ofUEqn.diag(),
--- a/test/test_pressureVelocityCoupling.cpp
+++ b/test/test_pressureVelocityCoupling.cpp
@@ -49,6 +49,7 @@ TEST_CASE("PressureVelocityCoupling")
         ),
         fvc::flux(ofU)
     );
+    ofPhi.correctBoundaryConditions();
 
     Foam::surfaceScalarField ofNu(
         Foam::IOobject(
@@ -61,6 +62,7 @@ TEST_CASE("PressureVelocityCoupling")
         mesh,
         Foam::dimensionedScalar("nu", Foam::dimensionSet(0, 2, -1, 0, 0), 0.01)
     );
+    ofNu.correctBoundaryConditions();
 
     auto nfPhi = NeoFOAM::constructFrom(rt.exec, rt.nfMesh, ofPhi);
     auto nfNu = NeoFOAM::constructFrom(rt.exec, rt.nfMesh, ofNu);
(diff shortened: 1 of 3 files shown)


test/test_pressureVelocityCoupling.cpp
```diff
@@ -86,6 +86,14 @@ TEST_CASE("PressureVelocityCoupling")
 
         Foam::volScalarField forAU("rAU", 1.0 / ofUEqn.A());
         nfUEqn.assemble();
+
+        nf::compare(
+            NeoN::la::upper(nfUEqn.linearSystem().matrix()),
+            ofUEqn.upper(),
+            ApproxVector(1e-15)
+        );
+        nf::compare(nfUEqn.linearSystem().matrix().diag(), ofUEqn.diag(), ApproxVector(1e-15));
```
greole: @HenningScheufler, @HendriceH  any idea why this test fails?

greole: I guess this is what 374a198852dabfb2ee4c938635484997831634e9 fixes.

greole: potentially related PRs [436](https://github.com/exasim-project/NeoN/pull/436) and [204 ](https://github.com/exasim-project/NeoFOAM/pull/204)

greole: Ok, I guess I was stupid and we discussed this several times already, but just for reference: we include            the boundary contributions to the diagonal elements in the Laplacian and div kernel already (`Kokkos::atomic_sub(&values[rowOwnStart + diagOffs[own]], valueMat);` ), but OpenFOAM doesn't.

test/test_pressureVelocityCoupling.cpp
```diff
@@ -86,6 +88,19 @@ TEST_CASE("PressureVelocityCoupling")
...
+            NeoN::la::upper(nfUEqn.linearSystem().matrix()),
+            ofUEqn.upper(),
+            ApproxVector(1e-15)
+        );
+
+        nf::compare(
+            NeoN::la::removeBoundaryContributions(nfUEqn.linearSystem()).matrix().diag(),
+            ofUEqn.diag(),
+            ApproxVector(1e-15)
+        );
```
chihtaw, line 105: The `removeBoundaryContributions` dependency should be documented in the test. The test implicitly encodes an important implementation detail:
```
NeoN includes BC contributions in the diagonal
OpenFOAM does not
```

I would recommend adding a comment directly above the comparison.

```suggestion
// NeoN stores boundary diagonal contributions directly in the matrix, whereas 
// OpenFOAM keeps them separate. Remove them before comparing against OpenFOAM 
// coefficients.
        nf::compare(
            NeoN::la::removeBoundaryContributions(nfUEqn.linearSystem()).matrix().diag(),
            ofUEqn.diag(),
            ApproxVector(1e-15)
        );
```

test/test_pressureVelocityCoupling.cpp
```diff
@@ -86,6 +88,19 @@ TEST_CASE("PressureVelocityCoupling")
 
         Foam::volScalarField forAU("rAU", 1.0 / ofUEqn.A());
         nfUEqn.assemble();
+
+        nf::compare(
+            NeoN::la::upper(nfUEqn.linearSystem().matrix()),
+            ofUEqn.upper(),
+            ApproxVector(1e-15)
```
chihtaw, line 95: The current tolerance may be overly strict on GPU backends.
This is probably fine on CPU/deterministic runs, but may become fragile on GPU. Consider something closer to 1e-12.

chihtaw (approved):
LGTM!
The only two things that I recommend before merge is adding explanatory comments around `removeBoundaryContributions` usage in the test and considering a tolerance which is not too strict for GPU backends.

PR #279, opened 2026-05-13
LLM based coding documentation

greole:
This PR adds a first  AGENTS.md  and instructions for LLM based coding. I think we can gradually extend the AGENTS.md file.

Additionally, to avoid flooding the repo with tool specific stuff CLAUDE.md and .planning are added to gitignore.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,95 @@
+# AGENTS.md
+
+This file provides guidance when working with code in this repository.
+
+## What is NeoFOAM
+
+NeoFOAM bridges **OpenFOAM** (the CFD framework) and **NeoN** (a GPU-portable computational backend built on Kokkos). It provides bidirectional data-structure converters, finite-volume CFD algorithms, and example solvers that run on CPUs and accelerator devices with the same source code.
+
+## Prerequisites
+
+OpenFOAM 2406+ must be sourced in the shell before any CMake commands:
+```sh
+source /path/to/OpenFOAM-2406/etc/bashrc
+```
+
+## Build Commands
+
+```sh
+cmake --list-presets                # show available presets
+cmake --preset develop              # configure (debug + tests + examples)
+cmake --build --preset develop      # build
+cmake --preset production           # configure (release, no tests)
+cmake --build --preset production   # build
+```
+
+Presets write build artifacts to `build/<presetName>/`. Binaries go to `build/<preset>/bin/`, libraries to `build/<preset>/lib/`.
+
+NeoN is located automatically: if `src/NeoN/CMakeLists.txt` exists it is used directly; otherwise it is fetched via CPM. Override with `-DNEOFOAM_NEON_DIR=<path>`.
+
+## Running Tests
+
+Tests require the `develop` preset (sets `NEOFOAM_BUILD_TESTS=ON`).
+
+```sh
+# Run all tests
+ctest --preset develop
+
+# Run a single test binary directly (must be run from its setup directory)
+cd test/setup_pressureVelocityCoupling
+../../build/develop/bin/tests/neofoam_pressureVelocityCoupling
+
+# Pass Catch2 options — separate them from OpenFOAM args with "---"
+../../build/develop/bin/tests/neofoam_operators "[SomeTag]" --- -case .
+```
+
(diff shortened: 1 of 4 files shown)


HendriceH (approved):
LGTM!

AGENTS.md
```diff
@@ -0,0 +1,95 @@
...
+A complete PISO incompressible solver that demonstrates the canonical usage pattern:
+1. `createAdapterRunTime` → maps fvSchemes/fvSolution
+2. `constructAndRegister` all fields into `VectorCollection`
+3. Build `PDESolver<Vec3>` for momentum with `dsl::imp::ddt + div + laplacian`
+4. PISO loop: `UEqn.solve(rhs)` → `computeRAUandHByA` → build pressure equation → `pEqn.solve()` → `updateFaceVelocity` → `updateVelocity`
+
+### `test/`
+
+Each test file starts an OpenFOAM `Time`/mesh in `test_main.cpp`, then Catch2 test cases use `GENERATE(allAvailableExecutor())` to run the same case on every compiled executor (Serial, CPUExecutor for OpenMP/threads, GPUExecutor for CUDA/HIP/SYCL). Setup case files live in `test/setup_<name>/`.
+
```
chihtaw, line 98: Consider adding a section about GPU execution semantics/Kokkos constraints

Since NeoFOAM heavily relies on NeoN/Kokkos portability, we can add something like
```
## GPU considerations
- Device kernels must use NEON_LAMBDA
- Avoid host-only allocations/access insider kernels
- Prefer parallelFor abstractions over raw loops
- Explicit synchronization may be required before host reads ( NeoN::fence(exec) )
- Use .copyToHost() when validating vectors in tests
```

AGENTS.md
```diff
@@ -0,0 +1,95 @@
...
+A complete PISO incompressible solver that demonstrates the canonical usage pattern:
+1. `createAdapterRunTime` → maps fvSchemes/fvSolution
+2. `constructAndRegister` all fields into `VectorCollection`
+3. Build `PDESolver<Vec3>` for momentum with `dsl::imp::ddt + div + laplacian`
+4. PISO loop: `UEqn.solve(rhs)` → `computeRAUandHByA` → build pressure equation → `pEqn.solve()` → `updateFaceVelocity` → `updateVelocity`
+
+### `test/`
+
+Each test file starts an OpenFOAM `Time`/mesh in `test_main.cpp`, then Catch2 test cases use `GENERATE(allAvailableExecutor())` to run the same case on every compiled executor (Serial, CPUExecutor for OpenMP/threads, GPUExecutor for CUDA/HIP/SYCL). Setup case files live in `test/setup_<name>/`.
```
chihtaw, line 97: Consider documenting testing conventions more explicitly
```suggestion
Tests should generally validate behavior across all enabled executors (CPU/GPU) and avoid assumptions about deterministic floating-point ordering.

Each test file starts an OpenFOAM `Time`/mesh in `test_main.cpp`, then Catch2 test cases use `GENERATE(allAvailableExecutor())` to run the same case on every compiled executor (Serial, CPUExecutor for OpenMP/threads, GPUExecutor for CUDA/HIP/SYCL). Setup case files live in `test/setup_<name>/`.
```

README.md
```diff
@@ -33,3 +33,8 @@ The repository is structured in the following way:
 - *tests* demonstrating that NeoFOAM and OpenFOAM deliver identical results are provided by this repository in the test folder.
 - *examples* provides examples of how NeoFOAM can be used for writing applications
 - *tutorials* provides tutorial cases which can be run like typical OpenFOAM cases
+
+## Agentic coding
+
+We provide an `AGENTS.md` file to support LLM-based coding workflows.
+Please instruct tools such as Claude Code to reference this file, for example by prompting: “See `AGENTS.md` for shared project instructions.”
```
chihtaw: The current wording might unintentionally imply a Claude-specific workflow. It could be slightly broader.
```suggestion
Please instruct AI coding tools to reference this file, for example by prompting: “See `AGENTS.md` for shared project instructions.”
```

chihtaw (approved):
LGTM!
The only major thing I think is missing is explicit GPU execution guidance.

AGENTS.md
greole: Good point so far, no one is using Claude or anything else on a GPU node, but this will come at some point.

PR #301, opened 2026-05-26
Replace nf::compare with custom matcher.

greole:
This PR replaces the bespoke nf::compare/NeoFOAM::compare test helper with dedicated Catch2 matchers (EqualsInternal / EqualsBoundary) to make test assertions more idiomatic and granular (internal vs boundary comparisons).


HendriceH (approved):
Looks good, only one thought: do we need to set a fixed random seed? I think I remember there migh be an issue when different processors generate the random values and the seed is not fixed?!

HenningScheufler (approved):

PR #306, opened 2026-05-28
Added continuity error calculation

HendriceH:
TODO: Need to add test

### DIFF
--- a/include/NeoFOAM/auxiliary/continuityError.hpp
+++ b/include/NeoFOAM/auxiliary/continuityError.hpp
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2025 - 2026 NeoFOAM authors
+
+#pragma once
+
+#include "NeoN/NeoN.hpp"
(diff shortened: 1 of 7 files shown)


include/NeoFOAM/auxiliary/continuityError.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * OpenFOAM continuityErrs.H formulation:
+ *   sumLocal  = dt * volWeightedAverage( |div(phi)| )   (abs, MPI-reduced)
+ *   global    = dt * volWeightedAverage(  div(phi)  )   (signed, MPI-reduced)
+ *   cumulative accumulates global across all PISO iterations
+ *
+ * @param phi           Face-flux NeoN SurfaceField
+ * @param rt            NeoFOAM RunTime (provides exec, mesh, MPI env, dt)
+ * @param cumulative    Running cumulative error — caller must initialise to 0
+ */
+inline void reportContinuityError(
```
greole: One thing i would like to be changed is that a function that computes something shouldn't print. so ideally reportContinuityError should be split in to computeContinuityError and the logging part.

greole (approved):

PR #315, opened 2026-06-04
perf(pdeSolver): solve momentum predictor in place, drop full matrix …

HendriceH:
PDESolver::solve(SpatialOperator&&) deep-copied the entire owned momentum LinearSystem (its nnz x Vec3 matrix values plus the boundary and off-diagonal matrices) every momentum solve, purely to carry the extra -grad(p) rhs while keeping the owned ls_ matrix rhs-free for the subsequent rAU/HbyA (H excludes grad p). The matrix and sparsity are identical between the two systems; only the rhs differs.

Assemble ls_ once, apply the explicit/-grad(p) source to ls_.rhs() in place for the solve, then add it back so ls_ is left holding the rhs-free H-system the rAU/HbyA step expects. The Ginkgo/PETSc solve path takes the system rhs as const, so restoring by re-adding the source reproduces the previous behaviour without duplicating the matrix. Removes the now-unused copy helper assemble(SpatialOperator&&).

### DIFF
--- a/examples/neoIcoFoam/neoIcoFoam.cpp
+++ b/examples/neoIcoFoam/neoIcoFoam.cpp
@@ -126,21 +126,27 @@ int main(int argc, char* argv[])
                 // Update the pressure BCs to ensure flux consistency
                 // Foam::constrainPressure(p, U, phiHbyA, rAU);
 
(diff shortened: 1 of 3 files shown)


greole (approved):
LGTM!

PR #316, opened 2026-06-05
test: assert NeoN L1-scaled residual matches OpenFOAM solverPerformance

HendriceH:
Add test_residual.cpp: build a random fvm::laplacian system, capture OpenFOAM solverPerformance::initialResidual() as ground truth, build the same system via PDESolver, solve with NeoN's opt-in 'l1ScaledResidual' stopping criterion, and assert the reported initial residual matches OpenFOAM and that the criterion converges below the absolute tolerance.

(NeoN backend changes live in the src/NeoN submodule repo, which develop does not pin as a gitlink.)


test/test_residual.cpp
```diff
@@ -0,0 +1,131 @@
...
+             {"criteria",
+              NeoN::Dictionary {{{"iteration", 0}, {"absolute_residual_norm", 1e-9}}}}}
+        };
+        auto solverStr = NeoN::la::Solver(exec, solverConfigStr);
+        auto statsStr = solverStr.solve(nfPDE.linearSystem(), nfT.internalVector());
+        REQUIRE(statsStr.entries.size() >= 1);
+        REQUIRE(statsStr.entries[0].initResNorm == Catch::Approx(ofInitResidual).epsilon(1e-6).margin(1e-12));
+    }
+
+    SECTION("L1 stopping criterion converges below tolerance " + execName)
```
greole, line 103: Ideally we should add a test that assert that unpreconditioned CG converges to same scaled residual on the OpenFOAM and neofoam side.

greole (approved):
LGTM! But please also add comparisson to final residual after some iterations with unpreconditioned CG.

PR #326, opened 2026-06-15
fail with better error message

greole:
Improves boundary-condition reading by adding support for OpenFOAM’s uniformFixedValue BC and by failing fast with clearer errors when encountering unsupported patch types (instead of deferring to an opaque std::bad_function_call).


lupeterm (approved):
lgtm

PR #347, opened 2026-06-29
Fix NFS filesystem error of removing output directories in unit tests

chihtaw:
## What is it about
Fix the NFS (Network File System) errors when attempting to remove output directories in the unit test of force coefficients.

## Outcome
Restructure the unit test so that
- No more NFS errors 
- Make it obvious whether the failure occurred during comparison or cleanup

### DIFF
--- a/test/test_forceCoeffs.cpp
+++ b/test/test_forceCoeffs.cpp
@@ -78,16 +78,16 @@ struct OfForcesSetup
 
     explicit OfForcesSetup(Foam::fvMesh& mesh, const Foam::Time& runTime, Foam::scalar nu = 0.0)
         : zeroU(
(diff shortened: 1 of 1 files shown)


chihtaw:
Update: 
make cleanup at the start of a unit test required, but make the cleanup at the end easier to identify than default behavior of Catch2

greole (approved):
Please check if SECTIONs are executed concurrently and if that would cause any issues with removing the files.

chihtaw:
> Please check if SECTIONs are executed concurrently and if that would cause any issues with removing the files.

I checked Catch2 documentation. Sections in a test case run one after another. As to test cases, I didn't find anything about parallel execution of test cases.

PR #355, opened 2026-07-10
test: cross-backend operator parity suite (pybFoam vs neon, explicit + implicit)

HenningScheufler:
## Summary

Adds `test/operators/`, a cross-backend parity suite that evaluates the FV
operators both backends expose — **explicit** (`interpolate(T)`, `flux(U)`,
`grad(T)`, `div(phi,T)`, `div(phi,U)`, `laplacian(Gamma,T)`, `laplacian(Gamma,U)`)
and **implicit** (`div(phi,T)`, `div(phi,U)`, `laplacian(Gamma,T)`,
`laplacian(Gamma,U)`) — with **pybFoam** (`fvc`/`fvm`) and **neon** (NeoN DSL via
the `nfb` bindings) on identical inputs, and asserts the results match.

The tests read like the maths they check — one backend facade per side:

```python
pyb_res = pyb.fvc.div(phi, U, scheme=scheme)   # OpenFOAM fvc::div
nb_res  = nb.imp.div(phi, T, scheme=scheme)    # NeoN implicit div, matrix-applied
np.testing.assert_allclose(nb_res, pyb_res, rtol=..., atol=...)
```

## Matrix — mesh × operator × scheme × executor

| axis | values |
|---|---|
| meshes | `cartesian_nx5` (5×5), `cartesian_nx20` (20×20), `sheared` — all generated in-process at test time via `pybFoam.meshing` (no external OpenFOAM binaries, no snappyHexMesh) |
| div schemes | `Gauss linear`, `Gauss upwind`, `Gauss linearUpwind grad(X)` (grad/laplacian/interp have a single variant from the checked-in `system/fvSchemes`) |
| executors | `Serial` (rtol 1e-9) and `GPU` (rtol 1e-8), the latter skipped on hosts without a device |

## Explicit vs implicit parity

- **Explicit** — both backends run the operator directly and return the internal
  field; neon goes through `nfb.evaluate_explicit` (zero the result, resolve the
  scheme from the fvSchemes `Dictionary`, run `explicitOperation`).
- **Implicit** — both assemble the operator and apply the matrix to the current
  field: pybFoam builds the `fvm` matrix and returns `M & psi`; neon assembles the
  linear system and returns `(A·psi − b) / V`. In exact arithmetic both equal the
  explicit operator, so the same reference fields validate both paths.


test/operators/test_operator_parity.py
```diff
@@ -0,0 +1,97 @@
...
+            if comp == 2 and two_d and spec.z_atol is not None:
+                comp_atol = max(atol, spec.z_atol * max(scale, 1.0))
+            _check(ref[:, comp], cand[:, comp], rtol, comp_atol, f"component {comp}")
+    else:
+        assert cand.shape == ref.shape, f"shape mismatch: {cand.shape} vs {ref.shape}"
+        _check(ref, cand, rtol, atol, "internal field")
+
+
+@pytest.mark.parametrize("executor", EXECUTORS)
+@pytest.mark.parametrize(("mesh", "op", "scheme"), PARAMS)
```
HenningScheufler: dont parametersize the ops but create for each op a new file

PR #384, opened 2026-08-01
feat(framework): operation-declared extension points, applied to incompressibleFluid MRF/fvOptions

HenningScheufler:
## What

A framework feature that lets an **operation declare how it can be extended**: an `ExtensionPoint` naming an interface class with one no-op-default method per call site. Models register implementation factories with `@<model>.extends(<point>)`; the dependency resolver injects every **active** implementation as an `Extensions` container into parameters annotated `Annotated[Extensions[Iface], <point>]`. Activation is model detection — a factory runs iff its registering ModelSpec has a live runtime for the case (matched by spec identity). The container is deliberately just iterable: operations call the sites in explicit, order-visible loops.

Applied to `incompressibleFluid`: the pressure-velocity operations declare one interface per operation (`MomentumExtension`, `PressureExtension`, `MeshUpdateExtension`) and MRF / fvOptions become registered implementations, removing all `mrf_zones`/`fv_options` optional-parameter branching from the PIMPLE and SIMPLE algorithms.

## Why

MRF and fvOptions had grown `if x is not None` branches at 7+ sites inside `momentum`/`continuity`/`mesh_update`; every future optional physics (porosity, rotorDisk, NeoN-side MRF) would multiply them. With this seam a new model is one isolated implementation class plus a factory — the operations don't change.

## Verification

- 20-step runs of mixerVessel2D (MRF), simpleCar (fvOptions porosity) and pitzDaily are **byte-identical** to the pre-refactor solver (association order in the momentum sum and every call placement preserved).
- `test_mrf.py` / `test_fv_options.py` (hook physics for both algorithms) pass **byte-unchanged**; 25 new framework unit tests; `test/framework/` + `test/solver/incompressibleFluid/` green except the pre-existing `test_pitzDaily_steady_comparison` SIGFPE failure (reproduces identically at the base commit).
- mypy (strict), ruff, reuse, typos pre-commit hooks pass; sphinx builds the new how-to (`doc/how-to/extend-operations.rst`) warning-free.

## Notes for review

- 3 commits: framework feature, solver refactor, how-to doc.
- `incompressibleVoF` still uses the optional-param form (its terms differ: `DDt(rho, U)`, `fvOptions(rho, U)`); same refactor can follow.
- Consuming modules must not use `from __future__ import annotations` (documented in the how-to).

https://claude.ai/code/session_01NbCNMfmvtXtsZTiCrrcZsQ

### DIFF
--- a/doc/how-to/extend-operations.rst
+++ b/doc/how-to/extend-operations.rst
@@ -114,12 +114,15 @@ A model contributes one plain function per site it acts at, with
 <neofoam.framework.model.spec.ModelSpec.contributes>`) — the same
 decorator used for interfaces. Sites are reached as attributes of the
 extension (``momentum_extension.terms``), so two extensions can share a
-site name:
+site name. The contributions live next to the model spec they belong to —
+in ``neofoam/mrf.py`` / ``neofoam/fv_options.py``:
 
 .. code-block:: python
 
-    from neofoam.fv_options import fvOptions
-    from neofoam.mrf import mrf
+    from neofoam.solver.incompressibleFluid.models.pressure_velocity.extension import (
+        momentum_extension,
+        pressure_extension,
+    )
 
 
     @mrf.contributes(momentum_extension.terms)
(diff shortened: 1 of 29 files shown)


HenningScheufler:
API iteration: the momentum extension sites are now expression-style — the operation reads as the equation:

```python
ext.correct_boundary_velocity(U)
UEqn = fvVectorMatrix(
    fvm.ddt(U) + fvm.div(phi, U) + viscousStress.divDevReff(U) + ext.terms(U)
)
UEqn.relax()
ext.constrain(UEqn)
```

`MomentumExtension.sources()` is gone: a source is a term on native's RHS, so the fvOptions implementation returns `negated(fv_options(U))` and it folds in with `-`. With no active model, `+ ext.terms(U)` returns the sum object itself — no arithmetic at all. Parity sweep (mixerVessel2D, simpleCar, pitzDaily) reproduces the archived baselines byte-for-byte.

https://claude.ai/code/session_01NbCNMfmvtXtsZTiCrrcZsQ

PR #386, opened 2026-08-04
incompressibleFluidNeoN-fixes

HenningScheufler:
# incompressibleFluidNeoN fixes

Third slice of the parity campaign on `test/incompressibleFluid`, following
#385 (`incompressibleFluid` / `incompressibleVoF`): the NeoN-twin fixes and
their parity coverage, extracted onto `stack/python_arch` so they review in
isolation. The shared groundwork (C++ core, `mapSolverSettings`, wall
functions, typed coeffs configs) is already on the base via #385 — this PR
carries only what was still missing.

## Fixes

* **Executor selection** (`fix(neon)`): `create_fields` passed no executor to
  `create_adapter_run_time`, so every run silently took the binding's `Serial`
  default. The NeoN runtime now reads `NEOFOAM_EXECUTOR`
  (`Serial`/`CPU`/`GPU`/`default`) once per process and the init step builds
  that executor.
* **Failed GPU init diagnostics**: Kokkos brings up every compiled backend, so
  a CUDA-enabled build takes a CUDA context per rank even for a Serial
  executor; when that fails, the bare CUDA error code is restated with the
  requested executor, MPI rank, device inventory and the ways out
  (`KOKKOS_VISIBLE_DEVICES`, fewer ranks per device, a CUDA-off build).

## Parity coverage pinned by new tests

* **Control-block selection**: a pisoFoam case (`PISO` block, no `PIMPLE`)
  initializes — both `create_pimple_state` and `create_pressure_reference`
  pick the block the case ships; a SIMPLEC cavity covers `consistent yes`.
* **Regex `fvSolution` keys**: `"(U|k|epsilon)"` / `".*Final"` solver entries
  resolve per field instead of dying in the NeoN hash-map lookup.
* **pitzDaily steady vs `incompressibleFluid`**: `cellLimitedGradU` on/off,
  corrected `snGrad`, SIMPLEC, and regex-keyed relaxation variants.
* **C++ parity**: `mapFvSchemes` hands OpenFOAM's `bounded <scheme>` div
  prefix through to NeoN's `BoundedDiv` (new Catch test), and the
  `pimpleParity` reference loop relaxes `UEqn` like `pimpleFoam`'s `UEqn.H`.
* **Turbulence closures**: kEpsilon/kOmegaSST dilatation `SuSp` terms on a
  non-solenoidal `phi`, and the `bound()` floors on near-zero / sub-floor `k`.

## Verification

* `pytest test/solver/incompressibleFluidNeoN test/solver/test_neon_runtime.py
  test/turbulence` — **108 passed** in 10:02 (1 pre-existing skip:
  neoPimpleFoam C++ nondeterminism; 1 pre-existing xfail: SA-DDES native
  closure), on a fresh editable install against OpenFOAM v2406.
* pre-commit (ruff, mypy, clang-format, reuse) green on every commit.

Extracted from `test/incompressibleFluid` by three-way merge against
`stack/python_arch`, keeping the base's #385 refactors (typed coeffs configs,
Extension hooks, consolidated test fixtures) where the branches had diverged.


test/boundedDivScheme.cpp
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
HenningScheufler: this test should be done in python

test/solver/incompressibleFluidNeoN/models/pressure_velocity/test_control_block_selection.py
```diff
@@ -0,0 +1,115 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: 2026 NeoFOAM authors
+
+"""Control-block selection during ``incompressibleFluidNeoN`` initialization.
```
HenningScheufler, line 4: too long

src/bindings/mrf.cpp
```diff
@@ -0,0 +1,572 @@
+// SPDX-FileCopyrightText: 2026 NeoFOAM authors
+// SPDX-License-Identifier: GPL-3.0-or-later
```
greole, line 2:
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later

// TODO: Implement a python version of this.
```

